PR #16, opened 2021-10-12
Use exception handling instead of returning none in EchoService and StidService

vetlek:


tests/api/apis/schedule/test_start_echo_mission.py
```diff
@@ -43,12 +48,18 @@ def test_start_mission(
...
     mock_start,
     expected_output,
     expected_status_code,
 ):
-    mocker.patch.object(EchoService, "get_mission", return_value=mock_get_mission)
+    mocker.patch.object(
+        EchoService,
+        "get_mission",
+        return_value=mock_get_mission,
+        side_effect=mock_get_mission_side_effect,
```
nicholasdalhaug, line 61: Why do we add the side_effect = None?

src/isar/services/service_connections/echo/echo_service.py
```diff
@@ -180,23 +173,21 @@ def get_mission(self, mission_id: int) -> Optional[Mission]:
...
-                tag_name=tag_name
-            )
-
-            measurement_steps: Optional[
-                List[Union[TakeImage, TakeThermalImage]]
-            ] = self.create_measurement_steps(tag_name=tag_name)
+            try:
+                drive_step: Optional[DriveToPose] = self.create_drive_step(
+                    tag_name=tag_name
+                )
```
aestene, line 185: The `create_drive_step` can return None but only because it uses the old return None functionality rather than raise an exception. See line 103 in this file.

This function should also cast an exception rather than returning None.

aestene (changes requested):
LGTM.

The branch must be rebased on top of main before it can be merged.

src/isar/apis/schedule/start_echo_mission.py
```diff
@@ -8,7 +8,9 @@
 from isar.config import config
 from isar.models.communication.messages import StartMessage, StartMissionMessages
 from isar.models.mission import Mission
-from isar.services.service_connections.echo.echo_service import EchoServiceInterface
+from isar.services.service_connections.echo.echo_service import (
+    EchoServiceInterface,
+)
```
tsundvoll, line 13: Double check line length. Standard for our project has been 88, so we should be consistent.

tsundvoll (changes requested):

PR #17, opened 2021-10-12
Handle excpetions instead of returning none

knord94:
Specifically inside mission_reader and map_reader


src/isar/services/readers/base_reader.py
```diff
@@ -35,7 +35,7 @@ def dict_to_dataclass(
     ) -> Optional[Any]:
         if not is_dataclass(target_dataclass):
             logger.error(f"{target_dataclass} is not a dataclass")
-            return None
+            raise BaseReaderError
```
nicholasdalhaug, line 38: Put the log message in the exception

src/isar/services/readers/base_reader.py
```diff
@@ -47,13 +47,17 @@ def dict_to_dataclass(
...
         except MissingValueError as e:
             logger.error(f"A value for a required field is not provided {e}")
-            return None
+            raise BaseReaderError
         except ValueError as e:
             logger.error({e})
-            return None
+            raise BaseReaderError
```
nicholasdalhaug, line 56: Why not propagate the same error farther?

src/isar/services/readers/mission_reader.py
```diff
@@ -89,18 +87,15 @@ def list_predefined_missions(self) -> Optional[dict]:
 
     def get_mission_by_id(self, mission_id) -> Optional[Mission]:
         mission_list_dict = self.get_predefined_missions()
-        if mission_list_dict is None:
-            logger.error(f"Found no missions")
-            return None
         try:
             return mission_list_dict[mission_id]["mission"]
         except Exception as e:
             logger.error(f"Could not get mission : {mission_id} - does not exist {e}")
```
nicholasdalhaug, line 93: Should we remove this logging and log farther up?

tests/api/apis/supervisor/test_supervisor_routes.py
```diff
@@ -87,6 +91,7 @@ class TestSupervisorRoutes:
                 1,
                 True,
                 None,
+                MissionReaderError,
```
nicholasdalhaug, line 94: Could this parameter be called something representing what it is? Like "exception_or_none" or something? "expected_exception"?

src/isar/services/readers/map_reader.py
```diff
@@ -37,13 +35,14 @@ def get_predefined_map_configs(self) -> Optional[dict]:
         json_files = self.predefined_map_config_folder.glob("*.json")
         for file in json_files:
             path_to_file = self.predefined_map_config_folder.joinpath(file.name)
-            map_config: MapConfig = self.get_map_config(path_to_file)
-            if map_config is None:
+            try:
+                map_config: MapConfig = self.get_map_config(path_to_file)
+            except BaseReaderError:
```
nicholasdalhaug, line 40: Can we rather let this fail instead of catching?

src/isar/services/readers/map_reader.py
```diff
@@ -37,13 +35,14 @@ def get_predefined_map_configs(self) -> Optional[dict]:
...
                 logger.warning(
                     f"File {path_to_file.as_posix()}"
                     f" could not be parsed to a map_config"
                 )
-            elif map_config.map_name in map_configs:
+            if map_config.map_name in map_configs:
```
nicholasdalhaug, line 45: Why do we need to handle duplicate map names?

src/isar/services/readers/mission_reader.py
```diff
@@ -21,32 +21,32 @@ def __init__(
...
-                mission: Mission = self.get_mission(path_to_file)
-                if mission is None:
+                try:
+                    mission: Mission = self.get_mission(path_to_file)
+                except BaseReaderError:
                     logger.warning(
                         f"File {path_to_file.as_posix()} could not be parsed to a mission"
                     )
-                elif mission.mission_id in invalid_mission_ids:
+                    continue
```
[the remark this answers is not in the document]
knord94: We might want to run the rest of the missions?

src/isar/services/readers/map_reader.py
knord94: Same as the above, do we want it to still run with the remaining map configs if we added a faulty new one?

nicholasdalhaug: I would want to fail, I think. This is your decision. You could ask someone else as well if you are not sure. I can jjoin that discussion if wanted.

src/isar/services/readers/mission_reader.py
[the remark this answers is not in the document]
nicholasdalhaug: Since these are files that are made by us beforehand, I would rather we fail and fix it if the files do not exist.

src/isar/services/readers/mission_reader.py
```diff
@@ -68,12 +68,10 @@ def get_predefined_missions(self) -> Optional[dict]:
             return missions
         except Exception as e:
             logger.error(f"Error in reading of files {e}")
-            return None
+            raise MissionReaderError
```
nicholasdalhaug, line 71: Should you not remove this as well? Either have log statement in exception or remove the whole except? The same question other places.

aestene (changes requested):
The commit messages have three different spelling ways for the word `exception`. This should be rephrased such that the spelling is correct.

src/isar/apis/schedule/start_mission.py
```diff
@@ -65,8 +65,9 @@ def get(self):
             self.logger.error(message)
             return message, HTTPStatus.NOT_FOUND
 
-        mission: Mission = self.mission_reader.get_mission_by_id(mission_id)
-        if mission is None:
+        try:
+            mission: Mission = self.mission_reader.get_mission_by_id(mission_id)
+        except MissionReaderError:
             message = StartMissionMessages.mission_not_found()
             return message, HTTPStatus.NOT_FOUND
```
aestene, line 72: Since this is at the top level of the functionality I would log the error here. So `except MissionReaderError as e` and `self.logger.error(f"Failed to read mission\n {e}")`

tests/api/services/readers/test_mission_reader.py
```diff
@@ -21,26 +22,22 @@
     ],
 )
 def test_get_mission(mission_reader, mission_path, expected_output):
-    output = mission_reader.get_mission(mission_path)
+    try:
+        output = mission_reader.get_mission(mission_path)
+    except:
+        output = None
```
aestene, line 28:
```suggestion
    try:
        output = mission_reader.get_mission(mission_path)
    except Exception:
        output = None
```

src/isar/services/readers/base_reader.py
```diff
@@ -47,16 +46,17 @@ def dict_to_dataclass(
...
-            return None
+            raise BaseReaderError("A value for a required field is not provided") from e
         except ValueError as e:
-            logger.error({e})
-            return None
+            raise BaseReaderError from e
         except Exception as e:
-            logger.error(f"Datastruct does not match expected struct {e}")
-            return None
+            raise BaseReaderError("Datastruct does not match expected struct") from e
```
Christdej, line 58: Not a strong preference, but I would consider removing the `try: except:` altogether and let a highet level catch the error. I think that the Error and message we get from the `dacite`-package is sufficient.

src/isar/services/readers/map_reader.py
```diff
@@ -38,12 +36,7 @@ def get_predefined_map_configs(self) -> Optional[dict]:
...
-            if map_config is None:
-                logger.warning(
-                    f"File {path_to_file.as_posix()}"
-                    f" could not be parsed to a map_config"
-                )
-            elif map_config.map_name in map_configs:
+            if map_config.map_name in map_configs:
                 logger.warning(
                     f"Map name {map_config.map_name} : {path_to_file.as_posix()}"
                     f" already exist. Skipping both original and duplicate ..."
```
Christdej, line 42: In the bottom of this file there is a function:
`def get_map_config_by_name(...`
Think we can remove the try/except and the Optional part of the return typing

PR #19, opened 2021-10-12
Move slimm and blob service to new storage module + Fix tests

tsundvoll:
This PR fixes the tests so that they are not dependent on credentials to run.

It also prepares for a more modular storage, by abstracting away slimm and blob storage one level.


nicholasdalhaug (approved):
LGTM

PR #43, opened 2021-10-14
Use local as default environment

aeshub:
Previously the program would be unable to start if the environment
variable ENVIRONMENT was not set to a valid value. Now the default
value will be "local" but a warning is logged if the variable has not
been set in the environment.

### DIFF
--- a/src/isar/config/config.py
+++ b/src/isar/config/config.py
@@ -1,8 +1,10 @@
+import logging
 from configparser import ConfigParser
 from os import getenv
 from pathlib import Path
 
 from dotenv import load_dotenv
+
 from isar.config.configuration_error import ConfigurationError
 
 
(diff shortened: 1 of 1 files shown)


src/isar/config/config.py
```diff
@@ -1,17 +1,24 @@
...
     def __init__(self):
         load_dotenv()
         env = getenv("ENVIRONMENT")
+
         if not env:
-            raise ConfigurationError(f"Environment not set")
+            env = "local"
+            logging.warning(
+                "ENVIRONMENT has not been set.\n Defaulting to local environment."
+            )
```
tsundvoll, line 20: Not sure this needs to be a warning. Maybe logging.info is sufficient? Could be a valid use case to just rely on the local environment.

tsundvoll (changes requested):

PR #56, opened 2021-10-18
Supply injector modules based on configuration

### LINKED ISSUE #55: Supply injector with modules based on configuration
The injector object is currently created in `__init__.py` in the `create_app` function. This should be moved to the top level `main.py` where the injector is created based on configuration. For example for storage there will be two modules `LocalStorageModule` and `BlobStorageModule` which are chosen prior to the creation of the injector based on the configuration.

sondreo:
Closes #55.

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,9 +1,43 @@
+from threading import Thread
+from typing import List
+
+from injector import Injector, Module
+
 from isar import create_app
+from isar.config import config
+from isar.modules import modules
+from isar.state_machine.state_machine import main
+
+
+def get_injector_modules() -> List[Module]:
+    injector_modules: List[Module] = []
+
+    module_config: dict = dict(config.items("modules"))
+
+    for module_key, module in modules.items():
+
+        injector_modules.append(
+            module[
+                "default"
+                if not module_key in module_config
+                else module_config[module_key]
+            ]
+        )
+
+    return injector_modules
+
 
 if __name__ == "__main__":
-    app = create_app()
 
-    host = app.config["HOST"]
-    port = app.config["PORT"]
+    injector_modules: List[Module] = get_injector_modules()
+    injector: Injector = Injector(injector_modules)
+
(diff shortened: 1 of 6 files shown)


src/isar/modules.py
```diff
@@ -131,3 +131,16 @@ def provide_transform(self, map_config_reader: MapConfigReader) -> Transformatio
...
+
+modules: dict = {
+    "coordinate_module": {"default": CoordinateModule},
+    "queues_module": {"default": QueuesModule},
+    "reader_module": {"default": ReaderModule},
+    "request_handler_module": {"default": RequestHandlerModule},
+    "robot_module": {"default": RobotModule},
+    "service_module": {"local": ServiceModule, "echo": ServiceModule},
+    "state_machine_module": {"default": StateMachineModule},
+    "storage_module": {"local": StorageModule, "blob": StorageModule},
```
aestene, line 144:
```suggestion
    "service_module": {"default": ServiceModule, "local": ServiceModule, "echo": ServiceModule},
    "state_machine_module": {"default": StateMachineModule},
    "storage_module": {"default": ServiceModule, "local": StorageModule, "blob": StorageModule},
```

aestene: You should have defaults for the modules that may be interchanged. When they are added with multiple modules the `local` will be the default one for these two for example.

main.py
```diff
@@ -1,9 +1,43 @@
...
+    return injector_modules
+
 
 if __name__ == "__main__":
-    app = create_app()
 
-    host = app.config["HOST"]
-    port = app.config["PORT"]
+    injector_modules: List[Module] = get_injector_modules()
+    injector: Injector = Injector(injector_modules)
```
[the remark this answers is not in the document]
sondreo: Agreed, implemented.

src/isar/modules.py
sondreo: Implemented default selections for all modules.

PR #70, opened 2021-11-04
Run send mission in separate thread

vetlek:


Eddasol (approved):
LGTM

src/isar/state_machine/states/send.py
```diff
@@ -26,70 +40,84 @@ def start(self):
 
     def stop(self):
         self.send_failure_counter = 0
+        if self.send_thread:
+            self.send_thread.join()
```
oysand, line 44: Could we risk the code hanging on this step if something is holding up the thread?

oysand: Well, that is the purpose of join come to think on it... What I mean is that is there a case where the code could hang indefinitely, if so, should we implement some sort of timeout?

src/isar/state_machine/states/send.py
```diff
@@ -26,70 +40,84 @@ def start(self):
...
-                    f"Failed to send mission after {self.send_failure_counter_limit} attempts. Cancelling mission."
+            if not self.send_thread:
+                self.state_machine.status.current_mission_step = (
+                    self._get_current_mission()
                 )
-                next_state: States = States.Cancel
-                break
+                self.send_thread = threading.Thread(
+                    target=send_mission_thread,
+                    args=(
```
oysand, line 68: Could we list which input each of these would be, for instance: schedule_step_func=self.state_machine.robot.schedule_step?

src/isar/state_machine/states/send.py
```diff
@@ -26,70 +40,84 @@ def start(self):
...
-                    computed_joints
-                )
-        else:
-            send_success = False
+        return self.state_machine.status.mission_schedule.mission_steps[0]
+
 
-        if self.state_machine.should_send_status():
-            self.state_machine.send_status()
+def send_mission_thread(
```
oysand, line 112: Would it be better to have this function as part of the class?

oysand (changes requested):
See comments/questions

src/isar/services/utilities/threaded_request.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class ThreadedRequest:
+    def __init__(self, request_func: Any) -> None:
+        self._thread: Optional[Thread] = None
+        self._request_func: Any = request_func
+        self._output: Any = None
+        self._output_lock: Lock = Lock()
+
+    def start_thread(self, *args) -> bool:
```
oysand, line 13: Should we specify which args one can input?

src/isar/state_machine/states/send.py
```diff
@@ -26,70 +29,71 @@ def start(self):
...
-                next_state: States = States.Cancel
-                break
+                self.send_thread = ThreadedRequest(
+                    self.state_machine.robot.schedule_step
+                )
+                self.send_thread.start_thread(
+                    self.state_machine.status.current_mission_step
+                )
+
+            if not self.send_thread.is_thread_alive():
```
oysand, line 61: Handle thread specific requirements inside ThreadRequest. Could try try except?

src/isar/state_machine/states/send.py
```diff
@@ -26,70 +33,75 @@ def start(self):
...
+                    self.state_machine.status.current_mission_step
+                )
+            try:
+                (
+                    send_success,
+                    mission_instance_id,
+                    computed_joints,
+                ) = self.send_thread.get_output()
+            except ThreadedRequestNotFinishedError:
+                continue
```
oysand, line 71: Consider a short sleep?

Christdej (approved):

PR #71, opened 2021-11-04
Run mission status request in separate thread

vetlek:


src/isar/state_machine/states/monitor.py
```diff
@@ -12,6 +13,11 @@
     from isar.state_machine.state_machine import StateMachine
 
 
+@dataclass
+class MissionStatusThreadOutput:
```
oysand, line 17: Should this dataclass be placed in the models folder and imported from there?

src/isar/state_machine/states/monitor.py
```diff
@@ -30,37 +39,54 @@ def start(self):
...
-                break
-
             if self.state_machine.should_send_status():
                 self.state_machine.send_status()
-            time.sleep(self.state_machine.sleep_time)
+
+            if not self.mission_status_thread:
+                self.mission_status_thread = threading.Thread(
+                    target=mission_status_thread,
+                    args=(
```
oysand, line 60: Could we list which input each of these would be, for instance: mission_status_func=self.state_machine.robot.mission_status?

src/isar/state_machine/states/monitor.py
```diff
@@ -86,3 +112,12 @@ def _log_status(self, mission_status: MissionStatus):
                 current_step=self.state_machine.status.current_mission_step,
             )
         self.iteration_counter += 1
+
+
+def mission_status_thread(
```
oysand, line 117: Would it be better to have this function as part of the class?

src/isar/state_machine/states/monitor.py
```diff
@@ -30,37 +39,54 @@ def start(self):
 
     def stop(self):
         self.iteration_counter = 0
+        if self.mission_status_thread:
+            self.mission_status_thread.join()
```
oysand, line 43: Could we risk the code hanging on this step if something is holding up the thread?

oysand (dismissed):
See comments/questions, large overlap with PR #70

src/isar/state_machine/states/monitor.py
vetlek: The code will hang if something is holding up the thread. It is here to make sure we don't have any threads going when we leave the state and make sure everything is closed properly. Could possibly kill the thread in some way, but I think it is bad practice. There should be some sort of timeout in the request module to limit the time we hang here, so should be a limited time in the worst case

src/isar/state_machine/states/monitor.py
sondreo: Implemented.

oysand (approved):
LGTM

Issue #63, opened 2021-10-21
Add default map implementation


aestene:
The current default map implementation uses a very particular map that will not make sense to new users. Suggestion is to add a default map where the robot and asset frame have the same base coordinate system. Thus there will be a simple default implementation.

A new map must be added to [this folder](https://github.com/equinor/isar/tree/main/src/isar/config/maps).

The default map configuration variable [here](https://github.com/equinor/isar/blob/main/src/isar/config/default.ini) should be changed to use the new map.

autt:
Resolved by #79

PR #79, opened 2021-11-09
New default map

thorvaldj:
An attempt in resolving #63. Robot and asset reference points are now equal. Resulting in no conversion between the coordinate systems. The reference points are set to be the same z=0 plane, perhaps giving an easier visualisation for tests?

### DIFF
--- a/src/isar/config/maps/default_map.json
+++ b/src/isar/config/maps/default_map.json
@@ -1,0 +1,49 @@
+{
+  "map_name": "default_map",
+  "robot_reference_points": {
+    "points": [
+      {
+        "x": 0.0,
+        "y": 0.0,
+        "z": 0.0,
+        "frame": "robot"
+      },
+      {
+        "x": 50.0,
+        "y": 100.0,
+        "z": 0.0,
+        "frame": "robot"
+      },
+      {
+        "x": 100.0,
+        "y": 50.0,
+        "z": 0.0,
+        "frame": "robot"
+      }
+    ],
+    "frame": "robot"
+  },
+  "asset_reference_points": {
+    "points": [
+      {
+        "x": 0.0,
+        "y": 0.0,
+        "z": 0.0,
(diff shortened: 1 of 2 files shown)


aestene:
Looks like a good solution! Using `z=0` as reference is a good idea.

Note that the transformation running in the background will cause some numerical computation effects on the results. Although this is of little consequence, it may be we should implement the possibility of toggling the transformation on and off in the future rather than performing the unity transformation this way. However, that is an issue for tomorrow.
image

Great contribution!

src/isar/config/maps/default_map.json
```diff
@@ -0,0 +1,50 @@
...
+          "x": 100.0,
+          "y": 50.0,
+          "z": 0.0,
+          "frame": "asset"
+        }
+      ],
+      "frame": "asset"
+    }
+  }
+
```
aestene, line 50: Add newline at end of file.

aestene (changes requested):
One nitpicky comment, otherwise it looks good. Let me know when you want to move the PR from draft so it may be merged.

src/isar/config/maps/default_map.json
aestene: Or, seems there is a newline but there are some empty spaces there.

thorvaldj: Ah. True. Force pushed a fix now.

thorvaldj:
> Note that the transformation running in the background will cause some numerical computation effects on the results. Although this is of little consequence, it may be we should implement the possibility of toggling the transformation on and off in the future rather than performing the unity transformation this way. However, that is an issue for tomorrow.

Yes, if I understand this correctly. You could perhaps toggle the transformation on/off on whether the initial reference points are equal or not.

aestene (approved):
LGTM. Thanks for the contribution!

PR #85, opened 2021-11-16
Add password authentication to FastAPI

knord94:


README.md
```diff
@@ -131,21 +131,31 @@ storage = local
...
+The API has an option to inlcude user authentication. This can be set in the [default configuration](./src/isar/config/default.ini).
+
+
+```
+authentication_enabled = false
+authentication_enabled = true
+```
+
+By default, the `local` storage module is used and API authentication is disabled. If using Azure Blob Storage a set of environment variables must be
+available which gives access to an app registration that may use the storage account. Enabling API authentication also requires the same environment variables. The required variables are
```
autt, line 151:
```suggestion
By default, the `local` storage module is used and API authentication is disabled. If using Azure Blob Storage, a set of environment variables which gives access to an app registration that may use the storage account must be available. Enabling API authentication also requires the same environment variables. The required variables are
```

src/isar/apis/security/authentication.py
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        self.token_algorithm = token_algorithm
+        self.token_expire_minutes = token_expire_minutes
+        self.pwd_context = CryptContext(schemes=["bcrypt"], deprecated="auto")
+
+    def verify_password(self, plain_password, hashed_password):
+        return self.pwd_context.verify(plain_password, hashed_password)
+
+    def get_password_hash(self, password):
```
autt, line 70: Is this function in use?

knord94: It is not, good catch!

autt (approved):
Looks good :tada: You should consider squashing the commits before merging.

PR #87, opened 2021-11-18
Refactor state machine tests

### LINKED ISSUE #61: Tests sometime fails for test_data_offload
The data offload test fails now and then. This test is timing dependent as there are two threads running in parallel.

It can be that the test calls some functions which we believe are mocked but after recent changes are not mocked anymore. Thus it could take more time to finish the test.

Christdej:
Closes #61


tests/integration/state_machine/test_state_machine.py
```diff
@@ -1,88 +1,66 @@
...
-def start_state_machine_in_thread(injector) -> StateMachine:
-    state_machine_thread: Thread = Thread(target=main, args=[injector])
-    state_machine_thread.daemon = True
-    state_machine_thread.start()
-    state_machine: StateMachine = injector.get(StateMachine)
-    return state_machine
+class StateMachineThread(object):
+    def __init__(self, injector) -> None:
+        self.injector = injector
+        injector.binder.bind(RobotInterface, to=MockRobot())
```
aestene, line 21: With the new way of setting up injector modules the binder should no longer be necessary. However, it requires the implementation of the `MockRobotModule` similar to `MockStorageModule` [here](https://github.com/equinor/isar/blob/main/tests/test_modules.py).

Then you would set the injector to use `MockRobotModule` [here](https://github.com/equinor/isar/blob/affbb4ed3475c31b95b6c5be8fc6ad50d79471d0/tests/conftest.py#L44) instead of `RobotModule`.

aestene: Since the `RobotModule` should be mocked in all tests it would be fine to do this in the base `injector`. For advanced tests it's possible to then setup a tailored `injector` which can pick and choose what should be mocked.

This can be a new issue if you wish or you can add it as part of this PR.

tests/integration/state_machine/test_state_machine.py
```diff
@@ -1,88 +1,66 @@
...
     message, _ = scheduling_utilities.start_mission(mission=mission)
     assert message.started
 
-    time.sleep(1)
-    assert state_machine.status.current_state is States.Monitor
-
     mocker.patch.object(
         MockRobot, "mission_status", return_value=MissionStatus.Completed
     )
     time.sleep(1)
```
aestene, line 44: Since the `MockRobot` implementation [here](https://github.com/equinor/isar/blob/main/tests/test_utilities/mock_interface/mock_robot_interface.py) returns `MissionStatus.InProgress` as default because of these timing based tests.

Should the default mock value for `mission_status` be changed to `MissionStatus.Completed` now that we are no longer relying on the timing based tests?

aestene: That would remove the need for this patch.

I don't think there are other tests than the timing based ones that rely on this default value.

tests/integration/state_machine/test_state_machine.py
Christdej: Good catch, will fix!

tests/integration/state_machine/test_state_machine.py
Christdej: Yeah, I agreed. Will change to MissionStatus.Completed

tests/integration/state_machine/test_state_machine.py
```diff
@@ -94,20 +72,30 @@ def test_state_machine_with_successful_collection(injector, mocker):
     )
```
aestene, line 72: On line 69 the `mocker.patch.object(MockRobot, "schedule_step", return_value=(True, 1, None))` patches to what is already the default return value for the mock itself. This line should be possible to remove.

aestene: Also, note that the patch on line 70 may be removed if the previous comment on default value for `mission_status` is done.

tests/integration/state_machine/test_state_machine.py
```diff
@@ -94,20 +72,30 @@ def test_state_machine_with_successful_collection(injector, mocker):
...
 
     message, _ = scheduling_utilities.start_mission(mission=mission)
+    assert message.started
     time.sleep(1)
```
aestene, line 76: Can this sleep be removed?

Christdej: It can, but then I will have to make some other logic to assure that the state machine is done running. If I do the assert without waiting, the state machine is not always done running and expected transition log will not match the transition log on the state machine.

tests/integration/state_machine/test_state_machine.py
Christdej: Good catch, will remove

tests/integration/state_machine/test_state_machine.py
```diff
@@ -120,8 +108,21 @@ def test_state_machine_with_unsuccessful_collection(injector, mocker):
     mocker.patch.object(MockRobot, "get_inspection_references", return_value=None)
 
     message, _ = scheduling_utilities.start_mission(mission=mission)
```
aestene, line 110: Same comments on the patches above (line 104, 105)

autt:
Do these tests need to be in a separate `integration` folder, or can they be located together with the other state machine tests?

tests/integration/state_machine/test_state_machine.py
aestene: True, still two threads after all!

aestene (changes requested):
Suggested changes to patching and default mocks. Otherwise very good improvements.

aestene (approved):
LGTM

PR #88, opened 2021-11-18
AAD Authentication

knord94:


aestene:
Note that the commit message for "wip Implement AAD Authentication" should be updated.

aestene (approved):
LGTM.

Have tested with and without authentication using both OpenAPI docs and direct requests from postman.

PR #90, opened 2021-11-23
Add tests for API authentication

knord94:


tests/isar/apis/security/test_authentication.py
```diff
@@ -0,0 +1,69 @@
...
+    claims = {
+        "iss": "ISAR",
+        "iat": 1637588872,
+        "exp": 1669124872,
+        "aud": "test_client@example.com",
+        "sub": "user_impersonation@example.com",
+    }
+
+    key = "qwertyuiopasdfghjklzxcvbnm123456"
+    token = jwt.encode(claims=claims, key=key)
```
autt, line 18: Is there a reason why the claims and key mock are this detailed? 

It should be sufficient with something like 

```
claims = {}
key = "some_key"
```

autt (changes requested):
Looks good :tada: 

Please see the minor comments. Also, the commits should be squashed before merging.

autt (approved):

PR #97, opened 2021-11-24
Remove `mission_created_by` field from metadata

### LINKED ISSUE #94: Remove email field from metadata
The email field `mission_created_by` is propagated from the mission planner if the email field is available. 

As we have no need of this information currently we should remove it according to GDPR.

https://github.com/equinor/isar/blob/c2dc26922770f573c010d1ed8d490ae34cc6fd16/src/isar/storage/storage_service.py#L212

autt:
We have no need for this information. Removed according to GDPR.

Resolves #94


aestene (approved):
LGTM

PR #102, opened 2021-11-25
Make STID service independent of plant

aestene:
Previously a plant was hardcoded but now the plant is decided
dynamically from configuration.


Christdej (approved):

aestene:
Closes #102

PR #103, opened 2021-11-25
Update robot interface

vetlek:
Suggestion for updated robot interface based on the Robot API meeting 24.11.21


src/robot_interface/robot_interface.py
```diff
@@ -1,146 +1,214 @@
...
         """
         raise NotImplementedError
 
     @abstractmethod
-    def log_status(
-        self, mission_id: Any, mission_status: MissionStatus, current_step: Step
-    ) -> None:
-        """Logs the status of the executing mission that is being monitored.
+    def get_sensor_recordings(self, sensor: Sensor) -> Sequence[InspectionResult]:
+        """Get the recorded data from the given sensor.
```
aestene, line 124: Why do we need both of these methods? When will we retrieve results that are not due to an inspection step but we need to link directly to a sensor?

src/isar/models/mission.py
```diff
@@ -5,15 +5,13 @@
...
 from isar.models.mission_metadata.mission_metadata import MissionMetadata
-from robot_interface.models.inspection.inspection import Inspection
-from robot_interface.models.mission import STEPS
+from robot_interface.models.step import Step
 
 
 @dataclass
 class Mission:
-    mission_steps: List[STEPS]
+    mission_steps: List[Step]
```
aestene, line 13: I suspect this will break the local mission planner and how it reads missions from a file.

The dacite package which is used to read from json requires that the subclasses are specified, otherwise there is no way for it to infer which steps a mission could consist of.

src/robot_interface/models/sensor/sensor.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@dataclass
+class Microphone(Sensor):
+    name: str = "microphone"
+
+
+@dataclass
+class LIDAR(Sensor):
+    name: str = "lidar"
```
aestene, line 26: I'm still not seeing the real need of adding the sensor class. If it's because we would like to hide Steps from the robot itself we have introduced a new thing called Sensor which the robot needs to understand.

It is more intuitive than Step seen from the robots POV, however we can solve that by renaming Step maybe?

One reason could be if we intend to support multiple sensors of the same type, however I'm not certain how we would like ISAR to manage possible different image sensors for different robots.

tsundvoll:
Remember to update isar-robot when the interface is updated

src/robot_interface/models/sensor/sensor.py
Christdej: Think I agree with Arnt here, because we dont have any aditional information on the sensors other than the name string. I think we can wait creating sensor classes until we need more information on the sensors.

src/robot_interface/robot_interface.py
```diff
@@ -1,146 +1,214 @@
...
+        RobotException if function fails
         """
         raise NotImplementedError
 
     @abstractmethod
-    def mission_status(self, mission_id: Any) -> MissionStatus:
-        """Retrieves status of the current executing mission for the robot.
+    def step_status(self, step: Optional[Step] = None) -> StepStatus:
+        """Function that returns the status of the progress for
+        a given step.
```
sondreo, line 79: Are we interested in checking the status of any step of a mission?

src/robot_interface/models/step/step.py
```diff
@@ -104,4 +121,4 @@ class TakeThermalImage(InspectionStep):
     tag_id: Optional[str] = None
 
 
-STEPS = Union[DriveToPose, DockingProcedure, TakeImage, TakeThermalImage]
+STEPS = Union[ContinousInspectionStep, DockingProcedure, TakeImage, TakeThermalImage]
```
tsundvoll, line 124: We still need the DriveToPose dataclass; right?

src/robot_interface/robot_interface.py
```diff
@@ -1,146 +1,214 @@
...
+        ------
+        RobotException if function fails
         """
         raise NotImplementedError
 
     @abstractmethod
-    def mission_scheduled(self) -> bool:
-        """Determines if a mission is scheduled on the robot.
+    def schedule_step(self, step: Step) -> None:
+        """Schedules a step on the robot.
```
tsundvoll, line 55: Step is renamed to task

src/robot_interface/robot_interface.py
aestene: I would think of this as checking the status of the current step in a mission. Otherwise we will have to implement some way of storing the state of previous steps.

vetlek:
Replaced with #154

PR #104, opened 2021-11-25
Restructure configuration

### LINKED ISSUE #98: Structure and document default configuration
Go through and remove redundant configuration.

Restructure and rename the fields which are outdated.

Document through comments the meaning of each variable.

aestene:
Main points:
- All configuration variables in `default.ini` now have a description.
- All redundant variables have been removed.
- The number of config sections has been reduced.
- Variables have been renamed to better reflect their purpose.

Closes #98


Christdej (approved):

src/isar/config/default.ini
```diff
@@ -1,84 +1,134 @@
...
+# FastAPI host
+api_host = localhost
+
+# FastAPI port
+api_port = 3000
+
+[modules]
+# Determines which mission planner module is used by ISAR
+# Options: [local echo]
+# Selecting a different mission planner module to local may require certain access rights
```
Eddasol, line 40:
```suggestion
# Selecting a different mission planner module than local may require certain access rights
```

src/isar/config/default.ini
```diff
@@ -1,84 +1,134 @@
...
+mission_planner = local
+
+# Determines which storage module is used by ISAR
+# Options [local blob]
+# Selecting a different storage module to local may require certain access rights
```
Eddasol, line 45:
```suggestion
# Selecting a different storage module than local may require certain access rights
```

src/isar/config/default.ini
```diff
@@ -1,84 +1,134 @@
...
-
-[fastapi]
-env = development
-run_host = localhost
-run_port = 3000
+# Timeout in seconds for direct HTTP requests made through the RequestHandler
+request_timeout = 10
+
+# Timeout in seconds for retrieving messages on a queue from state machine to API
+queue_timeout = 10
```
tsundvoll, line 13: Could this comment be written more clearly?

This timeout is actually only used while checking if there is a message on the queue. If there is no message on the queue after the timeout passes, the check will return False.

```suggestion
# Timeout in seconds for checking whether there is a message on a queue in the state machine
queue_timeout = 10
```

Eddasol (approved):

PR #108, opened 2021-11-25
Remove redundant missions to only support default

### LINKED ISSUE #106: Remove redundant missions from predefined missions list
There are several missions in the [predefined missions folder](https://github.com/equinor/isar/tree/main/src/isar/config/predefined_missions) which are redundant. These should be removed to avoid clutter.

There is no need for more than one predefined mission as default.

aestene:
Closes #106


autt (approved):
LGTM

Eddasol (approved):

PR #109, opened 2021-11-25
Log state transitions over multiple lines

### LINKED ISSUE #99: Log the state transitions over multiple lines
The state transitions list that is logged at the end of a mission can be long if the max transitions log is set high. If this is the case the log should automatically log this over multiple lines instead of in one line.

https://github.com/equinor/isar/blob/f847871a37ccb71fa3dd26ed7109021ec1abb2cb/src/isar/state_machine/states/cancel.py#L56

sondreo:
Cutoff value for newline in logging of state transitions is kept as a local variable, in order not to clutter the configuration file.

Closes #99.

### DIFF
--- a/src/isar/state_machine/states/cancel.py
+++ b/src/isar/state_machine/states/cancel.py
@@ -53,4 +53,16 @@ class Cancel(State):
         self.state_machine.to_next_state(next_state)
 
     def stop(self):
-        self.logger.info(f"State transitions: {self.state_machine.transitions_list}")
+        self._log_state_transitions()
+
+    def _log_state_transitions(self):
+        state_transitions: str = ", ".join(
+            [
+                f"\n  {transition}" if (i + 1) % 10 == 0 else f"{transition}"
+                for i, transition in enumerate(
+                    list(self.state_machine.transitions_list)
+                )
+            ]
+        )
+
+        self.logger.info(f"State transitions:\n  {state_transitions}")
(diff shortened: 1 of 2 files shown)


src/isar/state_machine/states/cancel.py
```diff
@@ -53,4 +53,10 @@ def start(self):
         self.state_machine.to_next_state(next_state)
 
     def stop(self):
-        self.logger.info(f"State transitions: {self.state_machine.transitions_list}")
+        state_transitions_list: list = [
+            f"\n  {transition}" if (i + 1) % 10 == 0 else f"{transition}"
+            for i, transition in enumerate(list(self.state_machine.transitions_list))
+        ]
```
aestene, line 59:
```suggestion
        state_transitions: list = [
            f"\n  {transition}" if (i + 1) % 10 == 0 else f"{transition}"
            for i, transition in enumerate(list(self.state_machine.transitions_list))
        ]
```

src/isar/state_machine/states/cancel.py
```diff
@@ -53,4 +53,10 @@ def start(self):
...
+
+        state_transitions: str = ", ".join(state_transitions_list)
+        self.logger.info(f"State transitions:\n  {state_transitions}")
```
aestene, line 62: What about setting this as a separate function? `log_state_transitions`.

Then the stop function doesn't have to have complex arithmetics directly within it.

aestene (changes requested):
Some suggestions, otherwise very good.

src/isar/state_machine/states/cancel.py
sondreo: Good idea, will do.

aestene (approved):
LGTM

PR #110, opened 2021-11-25
Set unique mission id for local planner

vetlek:


src/isar/models/mission.py
```diff
@@ -26,3 +26,11 @@ def __post_init__(self) -> None:
...
         if self.mission_metadata is None:
             self.mission_metadata = MissionMetadata(mission_id=self.mission_id)
+
+    def set_unique_mission_id(self) -> None:
+        plant_short_name: str = config.get("metadata", "plant_short_name")
+        robot_id: str = config.get("metadata", "eqrobot_robot_id")
+        now: datetime = datetime.utcnow()
+        self.mission_id = (
+            f"{plant_short_name.upper()}{robot_id.upper()}{now.strftime('%d%m%Y%H%M')}"
+        )
```
aestene, line 36: Can this function also be used above in `__post_init__` instead of duplicating the code?

aestene: Might be you have to make it static.

aestene:
The test error is abit weird.

You have to rebase your local changes on top of upstream main. Then you will get the same error when running tests locally.

The error is caused by the new `set_unique_mission_id` function using the config variable `eqrobot_robot_id` which has been renamed in the newest changes in main. If you rebase and rename the variable to just `robot_id` it should work.

aestene (approved):
LGTM

PR #114, opened 2021-11-26
Add step.id 

### LINKED ISSUE #96: Add unique IDs to every step
Add a unique ID to every step in a mission. The ID can then be used to manage inspections within isar-robot such that ISAR itself only has to manage the step ID.

This will remove the need for `vendor_mission_id` where this is present.

autt:
Resolves #96 and #149

This PR changes the robot interface, and will require corresponding change in ISAR robot packages.

There are PRs for [isar-robot](https://github.com/equinor/isar-robot/pull/18) and [isar-turtlebot](https://github.com/equinor/isar-turtlebot/pull/31) ready for review.


src/isar/state_machine/states/monitor.py
```diff
@@ -91,14 +89,13 @@ def _run(self):
         self.state_machine.to_next_state(next_state)
 
     def _mission_finished(
-        self, mission_status: MissionStatus, instance_id: int
+        self,
+        mission_status: MissionStatus,
     ) -> bool:
```
tsundvoll, line 94: Function definition could go on one line

src/robot_interface/models/mission/task.py
```diff
@@ -68,7 +69,8 @@ class DriveToPose(MotionTask):
     """
 
     pose: Pose
-    task_name: Literal["drive_to_pose"] = "drive_to_pose"
+    name: Literal["drive_to_pose"] = "drive_to_pose"
+    id: UUID = field(default_factory=uuid4)
```
tsundvoll, line 73: Could the id attribute be a part of the Task dataclass, since all subclasses needs it?

src/robot_interface/robot_interface.py
```diff
@@ -45,12 +45,12 @@ def mission_scheduled(self) -> bool:
...
     def mission_status(self, mission_id: Any) -> MissionStatus:
-        """Retrieves status of the current executing mission for the robot.
+        """Retrieves status of the current executing mission task for the robot.
 
         Parameters
         ----------
-        mission_id : Any
-            Unique identifier for the mission which the status should be checked for
+        task_id : UUID
+            Unique identifier for the task which the status should be checked for
```
tsundvoll, line 53: Docstring is renamed, but the parameter in the function def is still called mission_id

tests/isar/services/readers/test_mission_reader.py
```diff
@@ -94,6 +94,6 @@ def test_thermal_image_task(mission_reader):
 
     assert isinstance(task, TakeThermalImage)
     assert hasattr(task, "target")
-    assert task.task_name == "take_thermal_image"
+    assert task.name == "take_thermal_image"
```
tsundvoll, line 97: Add check for task.id
```suggestion
    assert task.name == "take_thermal_image"
    assert hasattr(task, "id")
```

tests/mocks/status.py
```diff
@@ -17,7 +16,6 @@ def mock_status(
     scheduled_status = Status(
         mission_status=mission_status,
         mission_in_progress=mission_in_progress,
-        current_mission_instance_id=current_mission_instance_id,
```
tsundvoll, line 20: We should make sure that the mock contains an ID on the step. Is this done? If not, and the tests are not failing, we probably should extend the test to cover this.

src/isar/mission_planner/local_planner.py
```diff
@@ -56,7 +56,7 @@ def read_mission_from_file(mission_path: Path) -> Optional[Mission]:
             dataclass_dict=mission_dict,
             target_dataclass=Mission,
             cast_config=[Frame],
-            strict_config=True,
+            strict_config=False,
```
tsundvoll, line 59: Why is this changed?

tsundvoll (changes requested):
Great work, this simplifies a lot! A few suggestions in the comments.

src/robot_interface/robot_interface.py
```diff
@@ -92,15 +88,11 @@ def log_status(
         raise NotImplementedError
 
     @abstractmethod
-    def get_inspection_references(
-        self, vendor_mission_id: Any, current_task: Task
-    ) -> Sequence[Inspection]:
+    def get_inspection_references(self, current_task: Task) -> Sequence[Inspection]:
```
tsundvoll, line 91: Does it have to be the current task we want to get the inspection reference from? Could this instead be renamed to just task? This is maybe a separate PR; but still an interesting thought.
```suggestion
    def get_inspection_references(self, task: Task) -> Sequence[Inspection]:
```

src/isar/models/communication/status.py
```diff
@@ -10,7 +10,6 @@
 class Status:
     mission_status: Optional[MissionStatus]
     mission_in_progress: bool
-    current_mission_instance_id: Optional[int]
     current_mission_task: Optional[Task]
```
aestene, line 13: Can we rename this to simply `current_task` while we're at it? The concept of `current_task` should be unique in this context now.

src/isar/state_machine/states/monitor.py
```diff
@@ -91,14 +89,13 @@ def _run(self):
...
         if mission_status == MissionStatus.Unexpected:
-            self.logger.error(
-                f"Mission status on task {instance_id} returned and unexpected status string"
-            )
+            self.logger.error("Mission status returned an unexpected status string")
         elif mission_status == MissionStatus.Failed:
-            self.logger.warning(f"Mission instance {instance_id} failed...")
+            self.logger.warning("Mission failed...")
```
aestene, line 98: Strictly speaking the Mission didn't fail, just a Task within the mission so this log message is not correct.

Would recommend logging something like "Task failed. Task ID: id" or something along those lines.

aestene: We discussed renaming `MissionStatus` to `TaskStatus` as it's the status of a single `Task` and not the mission as a whole. If you wish to include that here you may, otherwise I can create a new issue to do so.

src/robot_interface/models/mission/task.py
aestene: This would be preferrable, however dataclasses in python have a weird limitation here.

If you define a field with a default value in a parent dataclass the inheriting dataclass cannot have any fields without default values. This is because the inheritance for dataclasses "stacks" fields in such a way that the default field then comes before the non-default field.

aestene: I think it would be solved by using regular classes where the behavior might be a bit different if I remember correctly. Although that's not really a solution.

The same limitation is why `tag_id` is present in both `TakeImage` and `TakeThermalImage` but not in `InspectionTask`.

tests/mocks/status.py
autt: There is not an ID field in Status anymore. The id can be found in current_task.id.

src/robot_interface/robot_interface.py
aestene: Good idea. It would change the requirements to the implementation of the robot though depending on what the robot does.

For example, would we expect to retrieve inspections for "any" task id? I would assume just for the task IDs in the current mission.

src/isar/mission_planner/local_planner.py
autt: Reverted

src/robot_interface/models/mission/task.py
autt: I agree with @aestene here. I suggest that we keep this implementation for now, but looking at ways to improve the dataclass implmentation is left for issue #133.

src/isar/state_machine/states/monitor.py
autt: In general, the term `mission` is not used consistently and does not have a unique meaning in ISAR. Renaming is in process and will be added to this PR. `MissionStatus ` -> `TaskStatus` is among those changes.

src/robot_interface/robot_interface.py
autt: Interesting idea. It is now left for #147

aestene (approved):
LGTM. 
I would await merging until the implementation for all robot repos are ready in this case.

src/robot_interface/models/mission/task.py
```diff
@@ -79,6 +81,7 @@ class DockingProcedure(MotionTask):
 
     behavior: Literal["dock, undock"]
     name: Literal["docking_procedure"] = "docking_procedure"
+    id: UUID = field(default_factory=uuid4)
```
vetlek, line 84: Instead of setting `id` for all step classes, you can add `id: UUID = field(default_factory=uuid4, init=false)` in the base class

vetlek (changes requested):

src/robot_interface/models/mission/task.py
autt: Good catch :+1:

vetlek (approved):
LGTM

tsundvoll (approved):

aestene (approved):

PR #127, opened 2021-12-01
Add issue templates

aestene:


nicholasdalhaug (approved):
We should consider if we want to remove some of these if they are not used. We do not want a lot of content to maintain unnecessarily.

aestene:
We discussed this in the retrospective today and decided to try with these three. It may be that `improvement` and `feature` has a certain overlap but we'll see how it develops.

The goal is to force ourselves to provide more of a description when creating an issue.

PR #129, opened 2021-12-01
Auto assign new issues and PRs to project

aestene:


aestene:
The solution was not fit for the beta version of Github Projects.

PR #130, opened 2021-12-01
Change name from step to task

Christdej:
As the "step"-concept is becoming larger we need a broader name. "Task" was chosen


aestene (approved):
LGTM

PR #138, opened 2021-12-03
Add optional argument to wrap orientation angle

sondreo:


vetlek (approved):

Christdej (approved):

src/robot_interface/models/geometry/orientation.py
```diff
@@ -36,7 +36,11 @@ def __eq__(self, other):
             return True
         return False
 
-    def as_euler(self, degrees: bool = False) -> list:
+    def as_euler(
+        self,
+        degrees: bool = False,
+        wrap_angles: bool = False,
```
nicholasdalhaug, line 42: This should not be in ISAR if it is specific to a robot.

Also, I do not see why this should be a flag in a function.

Christdej: Do you also think "degrees: bool = False" should be a flag?

PR #150, opened 2021-12-06
Default `send_failure_counter_limit` in config

### LINKED ISSUE #128: Move send_failure_counter_limit to defaults.ini
Move number from 
[send.py](https://github.com/equinor/isar/blob/main/src/isar/state_machine/states/send.py)
Consider moving other "hardcoded" values aswell

autt:
Resolves #128


aestene (approved):
LGTM

PR #154, opened 2021-12-07
Update robot interface

vetlek:


src/isar/state_machine/states/cancel.py
```diff
@@ -36,21 +41,21 @@ def start(self):
...
+                    + f"Task: {task}"
+                )
+                continue
+            for res in result:
+                res.metadata.tag_id = task.tag_id
+                self._transform_results_to_asset_frame(res.metadata.time_indexed_pose)
+                inspections.append(res)
+                self.storage_service.store(
+                    self.state_machine.current_mission.id,
+                    res,
```
sondreo, line 57:
```suggestion
            results: List[Inspection] = self.state_machine.robot.get_task_result(task)
            if not results:
                self.logger.warning(
                    f"Failed to upload inspection result as no results were received. "
                    + f"Task: {task}"
                )
                continue
            for result in results:
                result.metadata.tag_id = task.tag_id
                self._transform_results_to_asset_frame(result.metadata.time_indexed_pose)
                inspections.append(result)
                self.storage_service.store(
                    self.state_machine.current_mission.id,
                    result,
```

src/isar/state_machine/states/cancel.py
```diff
@@ -64,6 +69,21 @@ def start(self):
     def stop(self):
         self._log_state_transitions()
 
+    def _transform_results_to_asset_frame(
+        self, time_indexed_pose: Union[TimeIndexedPose, List[TimeIndexedPose]]
+    ):
```
Christdej, line 74:
```suggestion
    def _transform_results_to_asset_frame(
        self, time_indexed_pose: TimeIndexedPose
    ):
```
Can we make this function only take a `TimeIndexedPose` and rather check if it is a single or a list where this function is called?

src/robot_interface/robot_interface.py
```diff
@@ -11,129 +9,85 @@ class RobotInterface(metaclass=ABCMeta):
     """Interface to communicate with robots."""
 
     @abstractmethod
-    def schedule_task(self, task: Task) -> bool:
-        """Schedules a Task on the robot.
-
-        The method must adapt the standard mission to the
-        specific robots mission planning.
+    def initiate_task(self, task: Task) -> bool:
```
Christdej, line 12: Should we return `None` and rather throw exception than returning a `bool`?

src/robot_interface/robot_interface.py
```diff
@@ -11,129 +9,85 @@ class RobotInterface(metaclass=ABCMeta):
...
-        -------
-        bool
-            True on successful abort of mission, false otherwise.
         """
         raise NotImplementedError
 
     @abstractmethod
-    def log_status(self, task_status: TaskStatus, current_task: Task) -> None:
-        """Logs the status of the executing task that is being monitored.
+    def stop(self) -> bool:
```
Christdej, line 55: Description of what this method returns and what it actually returns does not match. Think we should return `None` and throw exception on failing

Christdej (changes requested):

src/robot_interface/robot_interface.py
```diff
@@ -11,129 +9,85 @@ class RobotInterface(metaclass=ABCMeta):
...
-        inspection : Inspection
-            Inspection references to be downloaded.
+        task : Task
 
         Returns
         -------
-        Inspection
-            Returns the downloaded inspection with data attached as bytes.
+        Sequence[InpsectionResult]
+            List containing all the inspection results connected to the given task.
```
sondreo, line 85:
```suggestion
        Sequence[Inspection]
            List containing all the inspections connected to the given task.
```

src/isar/state_machine/states/cancel.py
vetlek: Think the `Union[TimeIndexedPose, List[TimeIndexedPose]]` is the most accurate since we can get both based on the inspection type.

Christdej (approved):

PR #158, opened 2021-12-08
Remove unused method in `StartMissionMessages`

### LINKED ISSUE #140: Can `invalid_mission_id(mission_id: int) -> StartMessage` be removed?
This method does not seem to be in use in ISAR. If it is not in use in other ISAR related repositories, it should be removed.

Should not change any functionality, just cleanup unused code. 

Should not affect Threat Model.

autt:
Close #140

### DIFF
--- a/src/isar/models/communication/messages/start_message.py
+++ b/src/isar/models/communication/messages/start_message.py
@@ -14,47 +14,15 @@ class StartMissionMessages:
         return StartMessage(started=True, message="Mission started")
 
     @staticmethod
-    def bad_request() -> StartMessage:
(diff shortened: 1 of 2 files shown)


tsundvoll (changes requested):
Good. I see that several other of the methods in this file is not in use. Could we remove them in the same commit?

autt:
Good catch @tsundvoll. The ones you listed are now removed, together with a last method that were only used in tests.

vetlek (approved):
LGTM

tsundvoll (approved):
Very good :+1:

PR #163, opened 2021-12-09
Refactor inspections

### LINKED ISSUE #146: Review Mission, Task, Inspection hierarchy
Both `tasks` and  `inspections` is a part of a `mission`. Consider having a clearer `mission` -> `task` -> `inspection` hierarchy

image

This should not change the functionality.

This should not affect the Threat Model.

### LINKED ISSUE #153: Consolidate Inspection and InspectionResult
Instead of dealing with two different inspection objects where one is a reference and the other holds data we should use one which has a datafield that is `None` until data is downloaded.

It will simplify the existing setup and reduce technical debt. 

There should be no change.

aestene:
Closes #146 and closes #153


src/robot_interface/models/inspection/inspection.py
```diff
@@ -17,16 +17,52 @@ class InspectionMetadata(ABC):
...
 
 @dataclass
 class Inspection:
     id: Any
     metadata: InspectionMetadata
+    data: Optional[bytes] = field(default=None, init=False)
+
+
+@dataclass
+class Audio(Inspection):
```
vetlek, line 52: Should we have Audio and Video as Inspections when they are not supported as InspectionTasks?

vetlek:
Have you considered making each inspection belong to the corresponding task and remove the list of inspections in the mission class?

aestene:
> Have you considered making each inspection belong to the corresponding task and remove the list of inspections in the mission class?

Yes, this is why the PR was moved back to draft as I had forgotten I was doing that as well :)

src/isar/state_machine/states/cancel.py
```diff
@@ -43,7 +52,10 @@ def start(self):
                         + f"Inspection reference: {inspection}"
                     )
 
-            self.storage_service.store_metadata(self.state_machine.current_mission)
+            if inspections:
```
vetlek, line 55: Should be moved one identation to the left

src/isar/storage/storage_service.py
```diff
@@ -155,7 +152,9 @@ def inspection_metadata_to_dicts(
                 }
             ]
 
-    def store_metadata_for_mission(self, mission: Mission) -> None:
+    def store_metadata_for_mission(
+        self, mission: Mission, inspections: List[Inspection]
```
vetlek, line 156: Take in mission_metadata only instead of the whole mission?

src/isar/storage/storage_service.py
```diff
@@ -40,10 +40,7 @@ def store(
 
         self.storage.store(data=result.data, path=destination_path)
 
-    def store_metadata(
-        self,
-        mission: Mission,
-    ) -> None:
+    def store_metadata(self, mission: Mission, inspections: List[Inspection]) -> None:
```
vetlek, line 43: Take in the mission id and the mission metadata instead of taking in the whole mission?

vetlek (changes requested):

src/isar/state_machine/states/cancel.py
aestene: Good catch!

vetlek (approved):
LGTM

PR #165, opened 2021-12-09
Add getting inspection type from echo

vetlek:


src/isar/mission_planner/echo_planner.py
```diff
@@ -1,3 +1,4 @@
+from enum import Enum
 import logging
```
tsundvoll, line 2: Swap those lines to get the imports sorted

src/isar/mission_planner/echo_planner.py
```diff
@@ -89,6 +88,21 @@ def _mission_plan(self, mission_id: int) -> dict:
 
         return response.json()
 
+    def _create_inspection_tasks_from_sensor_types(
+        self, tag_name: str, sensors: List[str]
+    ) -> List[Union[TakeImage, TakeThermalImage]]:
+        measurement_tasks: List[Union[TakeImage, TakeThermalImage]] = []
```
tsundvoll, line 94: I think it makes sense to just use `InspectionTask` i type hints here
```suggestion
    ) -> List[InspectionTask]:
        measurement_tasks: List[InspectionTask] = []
```

src/isar/mission_planner/echo_planner.py
tsundvoll, line 2: Swap lines to get the imports sorted

src/isar/mission_planner/echo_planner.py
```diff
@@ -172,3 +150,8 @@ def _create_thermal_image_task(self, tag_name: str) -> TakeThermalImage:
         )
 
         return thermal_image_task
+
+
+class EchoInspectionNames(Enum):
+    TakeThermalImage: str = "ThermicPicture"
+    TakeImage: str = "Picture"
```
tsundvoll, line 157: I think it is cleaner to make it a frozen (immutable) dataclass
```suggestion

@dataclass(frozen=True)
class EchoInspectionNames():
    TakeThermalImage: str = "ThermicPicture"
    TakeImage: str = "Picture"
```

src/isar/mission_planner/echo_planner.py
```diff
@@ -89,6 +88,21 @@ def _mission_plan(self, mission_id: int) -> dict:
...
         return response.json()
 
+    def _create_inspection_tasks_from_sensor_types(
+        self, tag_name: str, sensors: List[str]
+    ) -> List[Union[TakeImage, TakeThermalImage]]:
+        measurement_tasks: List[Union[TakeImage, TakeThermalImage]] = []
+        for sensor in sensors:
+            if sensor == EchoInspectionNames.TakeImage.value:
+                measurement_tasks.append(self._create_image_task(tag_name=tag_name))
+            elif sensor == EchoInspectionNames.TakeThermalImage.value:
```
tsundvoll, line 98: Making the EchoInspectionNames a dataclass; one can remove the .value attributes

```suggestion
            if sensor == EchoInspectionNames.TakeImage:
                measurement_tasks.append(self._create_image_task(tag_name=tag_name))
            elif sensor == EchoInspectionNames.TakeThermalImage:
```

src/isar/mission_planner/echo_planner.py
```diff
@@ -54,20 +52,21 @@ def get_mission(self, mission_id: int) -> Mission:
 
             try:
                 drive_task: DriveToPose = self._create_drive_task(tag_name=tag_name)
+                sensors = [x["sensorTypeKey"] for x in tag["sensorTypes"]]
```
tsundvoll, line 55: For readability, there should be a better name name than `x`.
In general, the naming here isn't great. Consider renaming
`mission_tags` to `plan_items`
`tag` to `plan_item`
`x` to `sensor_type` or something more clear

The format the sensor type is stored on is not the best, so it is a bit hard to work with.

tsundvoll (changes requested):

src/isar/mission_planner/echo_planner.py
vetlek: Using InspectionTask instead of Union[...] gives a type hint error since mission.tasks is a List[TASKS].

src/isar/mission_planner/echo_planner.py
vetlek: Changed it to use a function for the mapping of tasks

src/isar/mission_planner/echo_planner.py
```diff
@@ -20,6 +21,7 @@
 from robot_interface.models.geometry.pose import Pose
 from robot_interface.models.geometry.position import Position
 from robot_interface.models.mission import DriveToPose, TakeImage, TakeThermalImage
+from robot_interface.models.mission.task import TASKS, InspectionTask
```
tsundvoll, line 24: Unused imports

src/isar/mission_planner/echo_planner.py
```diff
@@ -118,57 +147,12 @@ def _create_drive_task(self, tag_name: str) -> DriveToPose:
...
-    def _create_thermal_image_task(self, tag_name: str) -> TakeThermalImage:
-        tag_position_robot: Position = self._get_tag_position_robot(tag_name=tag_name)
-
-        thermal_image_task: TakeThermalImage = TakeThermalImage(
-            target=tag_position_robot,
-            tag_id=tag_name,
-        )
 
-        return thermal_image_task
+def echo_sensor_type_to_isar_inspection_task(
```
tsundvoll, line 151: We could make this function a member function of the class, and prefix it with '_' since it is only used internally
```suggestion
@staticmethod
def _echo_sensor_type_to_isar_inspection_task(
```

tsundvoll (changes requested):

src/isar/mission_planner/echo_planner.py
tsundvoll: Ok, I get the same when I try here. Let's just keep it like this for now then.

src/isar/mission_planner/echo_planner.py
```diff
@@ -1,16 +1,17 @@
+from dataclasses import dataclass
```
tsundvoll, line 1: Also unused import
```suggestion
```

tsundvoll (changes requested):

tsundvoll (approved):
Looks good to me :+1:

PR #172, opened 2021-12-14
Remove $ from commands in readme

oysand:


.gitignore
```diff
@@ -135,3 +135,4 @@ dmypy.json
 
 # Temporary result folder
 results
+temp_results
```
tsundvoll, line 138: Isar-turtlebot specific things should be in isar-turtlebot and not in isar, right?

oysand: Agreed, but do we have a good way to store the temporary results in isar-turtlebot when the application runs from isar?

aestene: I agree with @tsundvoll, although there is no easy solution for this really. As such, as its only a `.gitignore` entry I would let it slide.

aestene (changes requested):
The first commit has a `.` at the end which should be removed.

aestene:
Before I noticed this PR I had already removed the `&` from the README in #174. The `temp_results` folder is already present in the `.gitignore` aswell so I think we'll close this PR. Let me know if there is something I've missed @oysand.

PR #186, opened 2022-01-24
Update task dependencies

### LINKED ISSUE #166: Use task id to indicate dependencies in mission
The current implementation of dependencies uses the index of the task in a mission to indicate dependencies. Instead the unique id of each task should be used to manage the dependencies.

There will be requirements to find the correct task within the mission based on the id rather than by directly indexing the list.

Not affected.

vetlek:
Closes #166

### DIFF
--- a/src/isar/state_machine/state_machine.py
+++ b/src/isar/state_machine/state_machine.py
@@ -21,10 +21,10 @@ from isar.services.coordinates.transformation import Transformation
 from isar.state_machine.states import Cancel, Idle, Monitor, Off, Send
 from isar.state_machine.states_enum import States
 from isar.storage.storage_service import StorageService
+from robot_interface.models.exceptions import RobotException
 from robot_interface.models.mission.status import TaskStatus
 from robot_interface.models.mission.task import Task
 from robot_interface.robot_interface import RobotInterface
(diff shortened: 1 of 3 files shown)


src/robot_interface/models/mission/task.py
```diff
@@ -56,7 +56,9 @@ class InspectionTask(Task):
 
     inspections: List[Inspection] = field(default_factory=list, init=False)
 
-    pass
+    @staticmethod
+    def get_inspection_type():
+        return Inspection
```
aestene, line 61:
```suggestion
    def get_inspection_type(self):
        return type(self)
```

src/isar/state_machine/state_machine.py
```diff
@@ -4,6 +4,7 @@
 from copy import deepcopy
 from typing import Deque, Optional, Tuple
 
+from black import re
```
aestene, line 7: Should this be here?

aestene (changes requested):
One minor comment for a line that might be removed, otherwise good.

src/isar/state_machine/state_machine.py
vetlek: Nope

aestene (approved):
LGTM

PR #189, opened 2022-01-27
Upload in separate thread to multiple storage handlers

### LINKED ISSUE #73: Ingest inspection results to SLIMM directly through their API
Currently all the inspections results are uploaded on the end of the mission when we enter cancel. This can be problematic if the mission is very long and we want the results as fast as possible (no need to get all inspections at the same time). If the state machine crashes due to internal or external issues, the inspection results are potentially lost. A potential fix is to upload the results in the collect state.

aestene:
Storage may now be configured as a list specifying modules that should be used for storage. The state machine will send any collected inspection to a separate uploader thread which handles uploads. It will loop through all selected storage handlers and upload the inspection to each one.

In addition a new storage handler has been created to support the SLIMM ingestion API.

Closes #73


aestene:
Currently there is no implementation for backing up and retrying uploads if the first attempt fails. This will be left for another issue.

src/isar/state_machine/states/monitor.py
```diff
@@ -69,6 +71,23 @@ def _run(self):
...
+                    inspections: Sequence[
+                        Inspection
+                    ] = self.state_machine.robot.get_inspections(
+                        task=self.state_machine.current_task
+                    )
+
+                    mission_metadata: MissionMetadata = (
+                        self.state_machine.current_mission.metadata
+                    )
+                    for inspection in inspections:
```
oysand, line 84: If there are multiple inspections for the current task, should the metadata be the same for all of them?

src/isar/storage/blob_storage.py
```diff
@@ -23,44 +26,47 @@ def __init__(
...
             return BlobServiceClient.from_connection_string(
                 self.storage_connection_string
             )
         except Exception as e:
-            logging.error(f"Unable to retrieve blob service client. Error: {e}")
+            self.logger.error(f"Unable to retrieve blob service client. Error: {e}")
             raise e
 
-    def get_container_client(self) -> ContainerClient:
+    def _get_container_client(self) -> ContainerClient:
```
oysand, line 68: If _get_container_client depends on the blob_servivice_client, should it take it as an argument or call _get_blob_service_client itself (depending if something else use the blob_service_client?)

src/isar/storage/slimm_storage.py
```diff
@@ -0,0 +1,103 @@
...
+    @inject
+    def __init__(self, request_handler: RequestHandler) -> None:
+        self.request_handler: RequestHandler = request_handler
+        self.logger = logging.getLogger("uploader")
+
+        self.credentials: DefaultAzureCredential = (
+            AzureCredentials.get_azure_credentials()
+        )
+
+        self.client_id: str = config.get("service_connections", "slimm_client_id")
```
oysand, line 28: Is client_id or scope used for anything other than building the request_scope? If so, do they need to be attached to self?

oysand: If not*

src/isar/storage/slimm_storage.py
```diff
@@ -0,0 +1,103 @@
...
+
+        inspection_type: str = get_inspection_type(inspection=inspection)
+        filename: str = get_filename(
+            mission_id=metadata.mission_id,
+            inspection_type=inspection_type,
+            inspection_id=inspection.id,
+            file_type=inspection.metadata.file_type,
+        )
+
+        m: MultipartEncoder = MultipartEncoder(
```
oysand, line 47: Consider a more descriptive name for the variable?

src/isar/storage/slimm_storage.py
```diff
@@ -0,0 +1,103 @@
...
+            AzureCredentials.get_azure_credentials()
+        )
+
+        self.client_id: str = config.get("service_connections", "slimm_client_id")
+        self.scope: str = config.get("service_connections", "slimm_app_scope")
+        self.request_scope: str = f"{self.client_id}/{self.scope}"
+
+        self.url: str = config.get("service_connections", "slimm_api_url")
+
+    def store(self, inspection: Inspection, metadata: MissionMetadata):
```
oysand, line 34: The store function currently does a lot. Would it make sense to split it out into supporting functions? Something like: _convert_to_slimm_format and _upload_to_slim?

src/isar/storage/uploader.py
```diff
@@ -0,0 +1,36 @@
...
+        self.storage_handlers: List[StorageInterface] = storage_handlers
+
+        self.logger = logging.getLogger("uploader")
+
+    def run(self) -> None:
+        while True:
+            inspection: Inspection
+            mission_metadata: MissionMetadata
+            try:
+                inspection, mission_metadata = self.upload_queue.get(timeout=5)
```
oysand, line 24: Should the timeout be hardcoded here? Also, is 5 sufficient for the uploader? If there is a larger inspection, for instance a video, could we expect it to need more time?

src/isar/storage/uploader.py
```diff
@@ -0,0 +1,36 @@
...
+            except Empty:
+                continue
+
+            for storage_handler in self.storage_handlers:
```
oysand, line 28: Do we need to support multiple storage_handlers?

src/isar/storage/utilities.py
```diff
@@ -0,0 +1,96 @@
...
+    folder: Path = Path(str(metadata.mission_id))
+    filename: str = get_filename(
+        mission_id=metadata.mission_id,
+        inspection_type=inspection_type,
+        inspection_id=inspection.id,
+        file_type=inspection.metadata.file_type,
+    )
+    path: Path = folder.joinpath(filename)
+
+    metadata_filename: str = get_filename(
```
oysand, line 25: Would it be cleaner to get the base file name and then append filetype and _meta.json for filename and metadata_filename respectively?

src/isar/storage/utilities.py
```diff
@@ -0,0 +1,96 @@
...
+    inspection_type: str = get_inspection_type(inspection=inspection)
+
+    folder: Path = Path(str(metadata.mission_id))
+    filename: str = get_filename(
+        mission_id=metadata.mission_id,
+        inspection_type=inspection_type,
+        inspection_id=inspection.id,
+        file_type=inspection.metadata.file_type,
+    )
+    path: Path = folder.joinpath(filename)
```
oysand, line 23: for consistency should we call it inspection_path?

src/isar/storage/utilities.py
```diff
@@ -0,0 +1,96 @@
...
+from robot_interface.models.inspection.inspection import Image, Inspection, ThermalImage
+
+
+def construct_local_paths(
+    inspection: Inspection, metadata: MissionMetadata
+) -> Tuple[Path, Path]:
+    inspection_type: str = get_inspection_type(inspection=inspection)
+
+    folder: Path = Path(str(metadata.mission_id))
+    filename: str = get_filename(
```
oysand, line 17: for consistency should we call it inspection_filename?

src/isar/state_machine/states/monitor.py
aestene: The mission metadata should be the same. This metadata contains no task specific information but information that is common for the entire mission. As such, this metadata will be the same across tasks as well. However, the inspection object itself also has a metadata object which is specific to that individual inspection.

aestene: Note that having the same metadata across the mission like this is what links the inspections together in SLIMM.

src/isar/storage/blob_storage.py
aestene: Implemented

src/isar/storage/slimm_storage.py
aestene: Trying with `multiform_body`

src/isar/storage/slimm_storage.py
aestene: Done

src/isar/storage/uploader.py
aestene: I don't see a reason to have this as a configuration variable. It could almost be argued that the queue could be blocking in this thread although I haven't gone for that approach as we will be adding additional functionality to handle back-off uploads of those that may have failed previously. 

The timeout could have been any value in principle.

src/isar/storage/uploader.py
aestene: Yes. Primary reason being that it might be necessary for us to upload everything to our own storage as a master of records for the data the robot produces. Additionally, analysis in early phases will be a lot easier if we can trigger from our own blob storage.

For testing it is also useful to have the local storage option to be able to inspect quickly what was created.

oysand: Agreed, the option is good to have

src/isar/storage/utilities.py
aestene: Changed it so both of them use `filename` and added the file ending afterwards. Also changed so they have the exact same filename only differing file endings. So `_meta` is gone.

src/isar/storage/utilities.py
aestene: I don't think so as it is now the filename of both files.

src/isar/storage/utilities.py
oysand: Sounds good

src/isar/storage/utilities.py
oysand: Agreed

src/isar/config/default.ini
```diff
@@ -46,8 +46,9 @@ api_port = 3000
 # Selecting a different mission planner module than local may require certain access rights
 mission_planner = local
 
-# Determines which storage module is used by ISAR
-# Options [local blob]
+# Determines which storage modules are used by ISAR
+# Comma separated list of modules to use. Each module will be called when storing results from inspections
+# Options [local blob slimm]
 # Selecting a different storage module than local may require certain access rights
 storage = local
```
oysand, line 53: If one is to add more is it [local, some_other, another]? If so, should the default be [ local ] to make it clearer that one can add more in this manner? It is explained in the text above, but who reads that before trying?

aestene: It is not a list as the config parser doesn't support that. It has to be a comma separated string as `local, blob, slimm`.

Can rename it to be a comma separated string

oysand: Aha, then changing it to string would make sense

aestene: Done

src/isar/config/default.ini
```diff
@@ -131,7 +144,7 @@ vertical_reference_system = MSL
 
 # Rotational representations of reported results
 # Options [quaternion]
-sensor_carrier_orientation_reference_system = quaternion
+media_orientation_reference_system = quaternion
```
oysand, line 147: Is media the right word here? And is quternion a reference system?

oysand: I struggle to come up with a better name

aestene: While I agree it's not a perfect name it is the name specified by the data scheme from SLIMM which means it could be more confusing to actually change it :)

oysand: Aha, then it makes sense to keep it

src/isar/config/default.ini
oysand: I still see comma seperated list

oysand (approved):
LGTM

PR #195, opened 2022-02-02
Return mission id from start endpoint

### LINKED ISSUE #194: Return mission id as part of response from start_mission endpoint
The mission_instance_id of the started mission should be returned in the response to post-requests to the start mission endpoint if the mission was successfully started. 

The mission_id should be returned as a part of the response body. A description of the structure of the return body should be shown in the swagger docs.

No change to the treat model.

vetlek:
Closes #194

### DIFF
--- a/src/isar/apis/api.py
+++ b/src/isar/apis/api.py
@@ -1,4 +1,5 @@
 import logging
+from http import HTTPStatus
 from logging import Logger
 from typing import Union
 
@@ -10,6 +11,7 @@ from fastapi.routing import APIRouter
 from injector import inject
 from pydantic import AnyHttpUrl
 
+from isar.apis.models.models import StartResponse, StopResponse
 from isar.apis.schedule.drive_to import DriveTo
 from isar.apis.schedule.start_mission import StartMission
 from isar.apis.schedule.stop_mission import StopMission
@@ -97,6 +99,24 @@ class API:
             self.start_mission.post,
             methods=["POST"],
             dependencies=[authentication_dependency],
+            responses={
+                HTTPStatus.OK.value: {
+                    "description": "Mission succesfully started",
+                    "model": StartResponse,
+                },
+                HTTPStatus.NOT_FOUND.value: {
+                    "description": "Not found - Mission not found",
+                    "model": StartResponse,
+                },
+                HTTPStatus.CONFLICT.value: {
+                    "description": "Conflict - Mission already ongoing",
+                    "model": StartResponse,
+                },
+                HTTPStatus.REQUEST_TIMEOUT.value: {
+                    "description": "Timeout - Could not contact state machine",
+                    "model": StartResponse,
+                },
+            },
         )
 
         router.add_api_route(
@@ -104,6 +124,16 @@ class API:
             self.stop_mission.post,
             methods=["POST"],
             dependencies=[authentication_dependency],
+            responses={
+                HTTPStatus.OK.value: {
+                    "description": "Mission succesfully stopped",
+                    "model": StopResponse,
+                },
+                HTTPStatus.REQUEST_TIMEOUT.value: {
+                    "description": "Timeout - Could not contact state machine",
+                    "model": StopResponse,
+                },
+            },
         )
         router.add_api_route(
             "/schedule/drive-to",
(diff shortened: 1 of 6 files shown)


src/isar/apis/api.py
```diff
@@ -9,7 +10,9 @@
 from fastapi.routing import APIRouter
 from injector import inject
 from pydantic import AnyHttpUrl
+from pyexpat import model
```
tsundvoll, line 13: Unused import

src/isar/apis/models/models.py
```diff
@@ -0,0 +1,16 @@
+from selectors import BaseSelector
+from typing import Optional
+from uuid import UUID
```
tsundvoll, line 3: Unused imports
```suggestion
from typing import Optional
```

src/isar/apis/schedule/start_mission.py
```diff
@@ -1,11 +1,13 @@
 import logging
+import re
 from dataclasses import asdict
 from http import HTTPStatus
 
-from fastapi import Query
+from fastapi import Query, Response
 from fastapi.responses import JSONResponse
```
tsundvoll, line 7: Unused import `JSONResponse`

src/isar/apis/schedule/start_mission.py
```diff
@@ -1,11 +1,13 @@
 import logging
+import re
 from dataclasses import asdict
```
tsundvoll, line 3: Unused import `asdict`

src/isar/apis/schedule/start_mission.py
```diff
@@ -41,14 +44,23 @@ def post(
...
+        message, status_code = response_scheduler
+        response.status_code = status_code.value
+        if status_code == HTTPStatus.OK:
+            return StartResponse(
+                message=message.message,
+                started=message.started,
+                mission_id=str(mission.id),
+            )
+        else:
+            return StartResponse(message=message.message, started=message.started)
```
tsundvoll, line 66: Have another look at this section. I am a bit confused by some parts of it, see if it can be made a bit clearer.

src/isar/apis/schedule/start_mission.py
```diff
@@ -41,14 +44,23 @@ def post(
...
-            self.logger.info(response)
+            response_scheduler = self.scheduling_utilities.start_mission(
+                mission=mission
+            )
+            self.logger.info(response_scheduler)
 
-        message, status_code = response
-        return JSONResponse(content=asdict(message), status_code=status_code)
+        message, status_code = response_scheduler
+        response.status_code = status_code.value
```
tsundvoll, line 58: For example this line; I don't see where response is used after this, so why is the status_code attribute set here?

src/isar/apis/schedule/start_mission.py
tsundvoll, line 66: Here it looks like we return a StartResponse including the mission ID only if the mission is scheduled properly. Else we also return a StartResponse, but without the mission ID. I don't see that the `StartResponse.message` will reflect that the mission was not ready to start though, since `message` is set before line 50, where the `response_scheduler` is retrieved.

src/isar/apis/schedule/stop_mission.py
```diff
@@ -2,9 +2,11 @@
 from dataclasses import asdict
 from http import HTTPStatus
 
+from fastapi import Response
 from injector import inject
 from starlette.responses import JSONResponse
 
+from isar.apis.models import StopResponse
```
tsundvoll, line 9: Check the imports, some are unused

tsundvoll (changes requested):

src/isar/apis/schedule/start_mission.py
vetlek: This is the way to set the status code in the fastapi. If this is not set we will return 200 in all cases. This is due to using pydantic models, which gives better documentation in the swagger docs, as return type opposed to using a JSONResponse as previous. See [fastapi](https://fastapi.tiangolo.com/advanced/response-change-status-code/)

tsundvoll (approved):
LGTM :+1:

PR #196, opened 2022-02-03
Add option for publishing to MQTT broker

tsundvoll:


src/isar/services/service_connections/mqtt/mqtt_client.py
```diff
@@ -0,0 +1,69 @@
...
+        self.client.on_disconnect = self.on_disconnect
+
+        self.client.username_pw_set(username=username, password=password)
+        self.client.connect(host=host, port=port, keepalive=60)
+
+    def on_connect(self, client, userdata, flags, rc):
+        self.logger.info("Connection returned result: " + mqtt.connack_string(rc))
+
+    def on_disconnect(self, client, userdata, rc):
+        # Add backoff strategy, retry after 5, 10, 30, 60 sec for example
```
tsundvoll, line 64: Should we remove this comment and add it as an issue?

tsundvoll (commented):
Only one comment. Apart from that it looks good to me. Can't approve, since it is originally my PR :sweat_smile:

src/isar/services/service_connections/mqtt/mqtt_client.py
aestene: Yes, that's a good idea :)

src/isar/state_machine/states/send.py
```diff
@@ -85,6 +94,10 @@ def _run(self):
 
             if send_success:
                 self.state_machine.current_task.status = TaskStatus.Scheduled
+
+                if self.state_machine.mqtt_client:
+                    self.state_machine.publish_task_status()
+
```
aestene, line 100: Remove this

src/isar/services/service_connections/mqtt/mqtt_client.py
aestene: #198

src/isar/state_machine/states/send.py
```diff
@@ -34,9 +34,17 @@ def start(self):
         self.state_machine.update_current_task()
         self.logger.info(f"State: {self.state_machine.current_state}")
 
+        if self.state_machine.mqtt_client:
+            self.state_machine.publish_task_status()
+            self.state_machine.publish_mission()
```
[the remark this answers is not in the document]
aestene: Since the mock makes the `mqtt_client` variable not `None` this actually works out well already.

PR #200, opened 2022-02-09
Add additional logging

### LINKED ISSUE #184: Logging of robot information from ISAR
With the changes in the robot interface introduced in v1.3.0 the functionality for logging robot specific information was removed from the interface. For debugging purposes it might be desirable to log what is going on in the robot specific implementation. An improvement would be to extract robot specific information that can make the debugging process easier and log them on regular intervals from ISAR.

This might lead to breaking changes in the robot interface. 

Logging of the wrong type of information (GDPR, etc) might lead to changes in the Threat Model.

aestene:
Closes #184

### DIFF
--- a/src/isar/apis/schedule/start_mission.py
+++ b/src/isar/apis/schedule/start_mission.py
@@ -35,6 +35,7 @@ class StartMission:
             description="ID-number for predefined mission",
         ),
     ):
+        self.logger.info("Received request to start new mission")
 
         try:
             mission: Mission = self.mission_planner.get_mission(mission_id)
@@ -50,8 +51,9 @@ class StartMission:
             response.status_code = status_code_ready_start.value
             return StartResponse(message=message.message, started=message.started)
 
+        self.logger.info(f"Starting mission: {mission.id}")
         response_scheduler = self.scheduling_utilities.start_mission(mission=mission)
-        self.logger.info(response_scheduler)
+        self.logger.info(f"Received response from State Machine: {response_scheduler}")
 
         message_scheduler, status_code_scheduler = response_scheduler
         response.status_code = status_code_scheduler.value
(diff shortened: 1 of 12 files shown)


aestene:
Example of log from running `isar_robot` mission
image

src/isar/apis/schedule/start_mission.py
```diff
@@ -35,6 +35,7 @@ def post(
             description="ID-number for predefined mission",
         ),
     ):
+        self.logger.info(f"Received request to start new mission")
```
Christdej, line 38:
```suggestion
        self.logger.info("Received request to start new mission")
```

Christdej (changes requested):
Just a small comment, else looks good!

vetlek:
Do we need to have the whole UUID of the Task in the overview of the tasks? Should be enough to use the first part to uniquely identify the task.

src/isar/state_machine/states/send.py
```diff
@@ -95,10 +97,14 @@ def _run(self):
...
                 next_state = States.Monitor
+                self.logger.info(
+                    f"Successfully scheduled {type(self.state_machine.current_task).__name__} "
+                    f"task: {self.state_machine.current_task.id}"
+                )
                 break
             else:
                 self.send_failure_counter += 1
-                self.logger.info("sending failed #: " + str(self.send_failure_counter))
+                self.logger.info("Sending failed #: " + str(self.send_failure_counter))
```
vetlek, line 107: Use f-string instead

src/isar/apis/schedule/start_mission.py
aestene: Done

aestene:
> Do we need to have the whole UUID of the Task in the overview of the tasks? Should be enough to use the first part to uniquely identify the task.

Fixed.

src/isar/state_machine/states/send.py
aestene: Done

vetlek (approved):
LGTM

PR #203, opened 2022-02-11
Remove map transformations

oysand:


src/isar/mission_planner/echo_planner.py
```diff
@@ -123,20 +120,15 @@ def _get_robot_pose(self, tag_name: str) -> Pose:
         predefined_pose: Pose = predefined_poses[tag_name]
 
         if predefined_pose.frame is Frame.Robot:
+            # TODO: Should we raise an error and require workaround pose to be asset frame?
```
oysand, line 123: Should we raise and error and require the workaround poses to be asset frame?

aestene: Yes.

src/isar/services/coordinates/transformation.py
```diff
@@ -1,9 +1,8 @@
 import logging
 from typing import Optional
 
-from alitra import AlignFrames, FrameTransform, Point, Quaternion
+from alitra import AlignFrames, FrameTransform, MapConfig, Point, Quaternion
```
aestene, line 4: Why not remove this file entirely?

aestene (changes requested):
One comment on a file I think could be removed.

Would like to implement this for `isar-turtlebot` and check that it works.

aestene (approved):
LGTM

PR #214, opened 2022-02-22
Change stale period to 30 days for PRs

aestene:


.github/workflows/stale.yml
```diff
@@ -29,4 +29,4 @@ jobs:
         stale-pr-label: 'stale'
         close-issue-message: 'This issue has been closed automatically due to a lack of activity.'
         close-pr-message: 'This pull request has been closed automatically due to a lack of activity.'
-        days-before-stale: 30
+        days-before-pr-stale: 30
```
anetteu, line 32: What about `days-before-issue-stale`? Is the default 60 days?

aestene: Yes, default is 60 days. However, I felt 60 days was a bit much for pull requests.

aestene: And I intended to use `dats-before-pr-stale` in the first place but copied the wrong setting...

anetteu (approved):

PR #218, opened 2022-03-02
Add docker-compose for robot and turtlebot

### LINKED ISSUE #192: Deploy ISAR using docker
Create a dockerfile which may be used to deploy ISAR.

ISAR should be easy to deploy, also in collaboration with Flotilla. 
- Note that the connections to a robot must be handled through the docker network interface (map ports)

The deployment of ISAR should be robust and only the required ports should be opened through the docker network interface.

aestene:
Closes #192

### DIFF
--- a/docker-compose-turtlebot.yml
+++ b/docker-compose-turtlebot.yml
@@ -1,0 +1,33 @@
+version: '3.7'
+services:
+  isar:
+    build:
+      context: .
+      args:
+        ROBOT_REPOSITORY_CLONE_URL: https://github.com/equinor/isar-turtlebot.git
+    image: isar:latest
+    ports:
+      - '3000:3000'
+    environment:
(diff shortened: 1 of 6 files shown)


docker-compose-turtlebot.yml
```diff
@@ -0,0 +1,33 @@
+version: '3.7'
+services:
+  isar:
+    build:
+      context: .
+      args:
+        ROBOT_REPOSITORY_CLONE_URL: https://${GITHUB_ACCESS_TOKEN}@github.com/equinor/isar-turtlebot.git
```
oysand, line 7: Why do we need the github access token to clone isar-turtlebot?

aestene: Ahh, good point. The docker-compose file was first made to be one file that could work for all robots, even those in a private repository. However, it then changed to be individual files for each as anything else didn't really work out. 

I will remove it.

GodVenn (approved):
LGTM

PR #222, opened 2022-03-03
Improve exception handling for 'start_mission' endpoint

### LINKED ISSUE #209: Don't create mission tasks when position is not known
**Describe the bug**
When running a mission and loosing contact with STID, the mission should not be created. There is no need to drive to the pose if the pose is not known.

**To Reproduce**
Try to schedule a mission when STID is not connected.

**Expected behavior**
Expect no mission or a mission only with known poses.

**Screenshots**

GodVenn:
Closes #209

### DIFF
--- a/src/isar/mission_planner/echo_planner.py
+++ b/src/isar/mission_planner/echo_planner.py
@@ -4,6 +4,7 @@ from typing import Dict, List, Union
 from azure.identity import DefaultAzureCredential
 from injector import inject
 from requests import Response
+from requests.exceptions import RequestException
 
 from isar.config import config
 from isar.config.predefined_poses.predefined_poses import predefined_poses
@@ -11,6 +12,7 @@ from isar.mission_planner.mission_planner_interface import (
     MissionPlannerError,
     MissionPlannerInterface,
 )
+from isar.models.communication.messages import StartMissionMessages
 from isar.models.mission import Mission
 from isar.services.auth.azure_credentials import AzureCredentials
 from isar.services.service_connections.request_handler import RequestHandler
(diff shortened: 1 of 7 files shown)


src/isar/mission_planner/echo_planner.py
```diff
@@ -46,8 +48,9 @@ def get_mission(self, mission_id: int) -> Mission:
         try:
             plan_items: List[dict] = mission_plan["planItems"]
         except KeyError as e:
-            self.logger.error("Echo request body don't contain expected keys")
-            raise MissionPlannerError from e
+            msg: str = "Echo request body don't contain expected keys"
+            self.logger.error(msg)
+            raise MissionPlannerError(msg)
```
aestene, line 53:
```suggestion
            raise MissionPlannerError(msg) from e
```
Is this possible?

src/isar/mission_planner/echo_planner.py
```diff
@@ -69,13 +72,19 @@ def get_mission(self, mission_id: int) -> Mission:
...
                     tag_name=tag_name, sensors=sensors
                 )
-            except (ValueError, Exception) as e:
-                self.logger.error(e)
+            except ValueError as e:
+                self.logger.error(f"Failed to create task with exception message: '{str(e)}'")
+                continue
+            except RequestException as e:
+                self.logger.error(f"Failed to create task with exception message: '{str(e)}'")
                 continue
```
aestene, line 80:
```suggestion
            except (ValueError, RequestException) as e:
                self.logger.error(f"Failed to create task with exception message: '{str(e)}'")
                continue
```

aestene (changes requested):
Have a look at the comments. Also, ensure you run Black formatting.

PR #223, opened 2022-03-04
Add retry by exponential backoff to uploader

### LINKED ISSUE #190: Handle failed uploads through retries
If an inspection result is attempted uploaded but the upload fails the uploader class should handle a backlog of uploads which should be retried for the storage handlers that failed a number of times over a specific period before the result is discarded.

The uploader receives the inspection on the queue and attempts to upload through all storage handlers. For each storage handler that fails the inspection is saved to a backlog which will re-attempt uploading for that specific storage handler.

- Note that the upload should not run again for any storage handlers where the upload was previously successful

The uploader thread was introduced through #189 and this issue aims to extend the functionality.

The threat model should not be affected by this feature.

GodVenn:
Closes #190 
Unsure about best practice for creating "struct" class UploaderRetryItem. Should this be in a different file?

### DIFF
--- a/src/isar/storage/uploader.py
+++ b/src/isar/storage/uploader.py
@@ -1,18 +1,69 @@
 import logging
+from dataclasses import dataclass
+from datetime import datetime, timedelta
 from queue import Empty, Queue
 from typing import List
 
+from isar.config import config
 from isar.models.mission_metadata.mission_metadata import MissionMetadata
 from isar.storage.storage_interface import StorageException, StorageInterface
 from robot_interface.models.inspection.inspection import Inspection
 
 
+@dataclass
+class UploaderRetryItem:
+    inspection: Inspection
+    mission_metadata: MissionMetadata
+    storage_handler: StorageInterface
+    _retry_count: int
+    _next_retry_time: datetime = datetime.max
+
+    def increment_retry(self, max_wait_time: int):
+        self._retry_count += 1
+        seconds_until_retry: int = min(2**self._retry_count, max_wait_time)
+        self._next_retry_time = datetime.utcnow() + timedelta(
+            seconds=seconds_until_retry
+        )
+
+    def get_retry_count(self) -> int:
+        return self._retry_count
+
+    def retry_is_ready(self) -> bool:
+        return datetime.utcnow() >= self._next_retry_time
+
+    def seconds_until_retry(self) -> int:
+        return max(0, int((self._next_retry_time - datetime.utcnow()).total_seconds()))
+
+
 class Uploader:
     def __init__(
-        self, upload_queue: Queue, storage_handlers: List[StorageInterface]
+        self,
+        upload_queue: Queue,
+        storage_handlers: List[StorageInterface],
+        max_wait_time: int = config.getint("DEFAULT", "upload_failure_max_wait"),
+        max_retry_attempts: int = config.getint(
+            "DEFAULT", "upload_failure_attempts_limit"
+        ),
     ) -> None:
+        """Initializes the uploader.
+
+        Parameters
+        ----------
+        upload_queue : Queue
+            Queues used for cross-thread communication.
+        storage_handlers : List[StorageInterface]
+            List of handlers for different upload options
+        max_wait_time : float
+            The maximum wait time between two retries (exponential backoff)
+        max_retry_attempts : int
+            Maximum attempts to retry an upload when it fails
+
+        """
         self.upload_queue: Queue = upload_queue
         self.storage_handlers: List[StorageInterface] = storage_handlers
+        self.max_wait_time = max_wait_time
+        self.max_retry_attempts = max_retry_attempts
+        self.retry_queue: List[UploaderRetryItem]
 
         self.logger = logging.getLogger("uploader")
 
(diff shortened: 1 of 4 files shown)


GodVenn:
Fixing mypy now

src/isar/storage/uploader.py
```diff
@@ -22,21 +70,66 @@ def run(self) -> None:
...
+                )
+
+    @staticmethod
+    def _retry_is_ready(item: UploaderRetryItem) -> bool:
+        return datetime.utcnow() >= item.get_retry_time()
+
+    def _process_retry_queue(self):
+        ready_items: List[UploaderRetryItem] = list(
+            filter(lambda x: Uploader._retry_is_ready(x), self.retry_queue)
+        )
```
aestene, line 125:
```suggestion
        ready_items: List[UploaderRetryItem] = [x if Uploader._retry_is_ready(x) for x in self.retry_queue]
```
I'm not certain what is considered the best approach when it comes to using the filter function here, but this is an alternative in Python which is called a list comprehension. It should yield the same result I think and you don't have to cast to list etc.

src/isar/storage/uploader.py
```diff
@@ -22,21 +70,66 @@ def run(self) -> None:
...
+        self.retry_queue = list(
+            filter(lambda x: not self._retry_is_ready(x), self.retry_queue)
+        )
```
aestene, line 128: List comprehension may be used here as well in the same manner as above.

src/isar/storage/uploader.py
```diff
@@ -1,18 +1,66 @@
...
 from robot_interface.models.inspection.inspection import Inspection
 
 
+@dataclass
+class UploaderRetryItem:
+    inspection: Inspection
+    mission_metadata: MissionMetadata
+    storage_handler: StorageInterface
+    __retry_count: int
+    __next_retry_time: datetime = datetime.max
```
aestene, line 19:
```suggestion
    _retry_count: int
    _next_retry_time: datetime = datetime.max
```
Internal variables in Python are usually prefixed with one underscore.

aestene (changes requested):
Very good! I like the approach you have taken.

Address the comments. In addition it would be good if you wrote a test which verifies the functionality by mocking failed uploads and then retrying etc. Let me know and we can look at the approach for this together.

src/isar/storage/uploader.py
```diff
@@ -22,21 +70,66 @@ def run(self) -> None:
...
+                    f"Retrying in {max(0, (retry_item.get_retry_time() - datetime.utcnow()).total_seconds())}s."
+                )
+            else:
+                self.logger.error(
+                    f"Storage handler: {type(storage_handler).__name__} "
+                    f"exceeded max retries to upload inspection: {str(inspection.id)[:8]}. Aborting upload."
+                )
+
+    @staticmethod
+    def _retry_is_ready(item: UploaderRetryItem) -> bool:
```
vetlek, line 119: Make this a method on the UploaderRetryItem instead? Then we don't need to have the get_retry_time() method

src/isar/storage/uploader.py
```diff
@@ -22,21 +70,66 @@ def run(self) -> None:
             inspection: Inspection
             mission_metadata: MissionMetadata
             try:
-                inspection, mission_metadata = self.upload_queue.get(timeout=5)
+                inspection, mission_metadata = self.upload_queue.get(timeout=1)
             except Empty:
+                if self.retry_queue:
```
vetlek, line 75: Move this out of the except. I think we want to run  `self._process_retry_queue()` all the time and not only when there is no new inspections. This can potentially lead to a deadlock if we always have a new inspection ready when we check the queue

vetlek:
Might be that you considered this approach already.  Could we use a single uploader queue with UploaderRetryItem (or perhaps UploaderItem) where all new inspections are added to the same queue as the inspections that needs retries. Then in the run method we add all new inspections to the queue and have a single `process_upload_queue()` (or some similar name) where all inspections are processed (check if ready for a new retry) and are removed from when they are successfully uploaded or reach the maximum number of retries . I would also suggest that we try to avoid creating new UploadRetryItems after each retry, since the data in the inspection can potentially be very large (possible memory leak, don't know how good the garbage collector is).

GodVenn:
> Might be that you considered this approach already. Could we use a single uploader queue with UploaderRetryItem (or perhaps UploaderItem) where all new inspections are added to the same queue as the inspections that needs retries. Then in the run method we add all new inspections to the queue and have a single `process_upload_queue()` (or some similar name) where all inspections are processed (check if ready for a new retry) and are removed from when they are successfully uploaded or reach the maximum number of retries . I would also suggest that we try to avoid creating new UploadRetryItems after each retry, since the data in the inspection can potentially be very large (possible memory leak, don't know how good the garbage collector is).

Will look into this!

GodVenn:
> Very good! I like the approach you have taken.
> 
> Address the comments. In addition it would be good if you wrote a test which verifies the functionality by mocking failed uploads and then retrying etc. Let me know and we can look at the approach for this together.

Will add a test!

GodVenn:
> Might be that you considered this approach already. Could we use a single uploader queue with UploaderRetryItem (or perhaps UploaderItem) where all new inspections are added to the same queue as the inspections that needs retries. Then in the run method we add all new inspections to the queue and have a single `process_upload_queue()` (or some similar name) where all inspections are processed (check if ready for a new retry) and are removed from when they are successfully uploaded or reach the maximum number of retries . I would also suggest that we try to avoid creating new UploadRetryItems after each retry, since the data in the inspection can potentially be very large (possible memory leak, don't know how good the garbage collector is).

I remember that this was my original idea too, but with my approach then, the two "item types" didn't align very well, because the retryitems had several upload handlers as part of the item. I see now that after adding one item per handler, it is easy to join the two to a single queue item type, and I agree that this is cleaner, so I will do that next.
Regarding the issue of creating new UploadRetryItems after each retry, I was under the impression that the inspection items would not be deep copied, but just by reference, as their references are never deleted, but rather passed around. Is this wrong?

GodVenn:
commit for adding tests is not completed, but pushed to branch as safety because the work is done on a VM.

src/isar/storage/uploader.py
```diff
@@ -22,21 +72,48 @@ def run(self) -> None:
...
+                    f"Storage handler: {type(upload_item.storage_handler).__name__} "
+                    f"failed to upload inspection: {str(upload_item.inspection.id)[:8]}. "
+                    f"Retrying in {upload_item.seconds_until_retry()}s."
+                )
+            else:
+                self.logger.error(
+                    f"Storage handler: {type(upload_item.storage_handler).__name__} "
+                    f"exceeded max retries to upload inspection: {str(upload_item.inspection.id)[:8]}. "
+                    "Aborting upload."
+                )
```
aestene, line 112: If this statement is hit should the upload item be removed from the internal queue?

aestene (approved):
LGTM

PR #226, opened 2022-03-08
Bug failed task causes state machine crash

GodVenn:
Closes [Failed task caused the state machine to crash](https://github.com/equinor/isar_taurob/issues/81) in isar_taurob
I couldn't find any other states in the state machine that did not have exception handling for method calls to the RobotInterface, so I think this PR should cover that for now.


GodVenn:
Also, I did not change the "next_state" based on the task result, as that functionality was already there, but at first sight it makes sense to only set the next_state to "Send" if we are also getting the inspections, any thoughts?

vetlek (approved):
LGTM! One unused import

MortFred (approved):

Issue #207, opened 2022-02-17
Fix failing integration test for turtlebot


aestene:
**Describe the bug**
The test fails as we have split the uploader functionality into a separate thread and we have started using a new schema for the data format.

**To Reproduce**
Start the turtlebot simulator and run the test here: https://github.com/equinor/isar/blob/decb3e45d3a83d7134aa0731978801f1d2f5676b/tests/integration/turtlebot/test_successful_mission.py#L84

**Expected behavior**
The test should pass once the turtlebot has finished the mission.

**Screenshots**
![image](https://user-images.githubusercontent.com/14331473/154502232-4575a3cf-5776-496e-ae18-56407f670d32.png)

PR #230, opened 2022-03-10
Use pydantic settings management

### LINKED ISSUE #219: Use pydantic settings management for configuration
The current configuration solution is not the best for overwriting from environment variables and it requires reading from files every time the config class is imported. 

The pydantic settings class should be created through a module and injected through injector to ensure the same settings are propagated throughout the system without needing to re-read files.

Should not be affected. Keep in mind that all configuration variables can be overwritten through the environment.

aestene:
Closes #219 and #207


src/isar/config/settings.env
```diff
@@ -0,0 +1,9 @@
+ISAR_ROBOT_PACKAGE = isar_turtlebot
+
```
GodVenn, line 2: Shouldn't this be defaulted to 'isar_robot' ?

aestene: Correct, good catch!

GodVenn (approved):
LGTM

PR #234, opened 2022-03-30
Mqtt connection robustness

aestene:


src/isar/config/settings.py
```diff
@@ -189,11 +189,11 @@ class Settings(BaseSettings):
...
     MISSION_TYPE: str = Field(default="inspection")
 
-    ISAR_STATE: str = Field(default=f"{ROBOT_ID}/isar_state")
+    ISAR_STATE: str = Field(default=f"{ROBOT_ID.default}/isar_state")
 
-    ISAR_MISSION: str = Field(default=f"{ROBOT_ID}/isar_mission")
+    ISAR_MISSION: str = Field(default=f"{ROBOT_ID.default}/isar_mission")
 
-    ISAR_TASK_STATUS: str = Field(default=f"{ROBOT_ID}/isar_task_status")
+    ISAR_TASK_STATUS: str = Field(default=f"{ROBOT_ID.default}/isar_task_status")
```
Christdej, line 196: Why `.default?`

Christdej (changes requested):

src/isar/config/settings.py
aestene: I found a better solution now, the previous one probably didn't even work if you wanted to change `ROBOT_ID`.

The reason was that just `ROBOT_ID` was not using the string value but adding a string representation of the `field` object so the topic would be wrong. This is now resolved using a validator instead.

Christdej (approved):

PR #241, opened 2022-04-11
Add validation of robot capabilities before starting mission

### LINKED ISSUE #134: Add validation of missions compatibility to robot
ISAR should verify that a mission is compatible with the robot that ISAR is monitoring prior to the mission being sent to the state machine for execution. Validation implies checking that the robot implementation supports all the Tasks which the mission consists of.

It is likely that this should be implemented with #95 such that ISAR may retrieve which Tasks the robot can perform.

No immediate effect on the current threat model.

vetlek:
Closes #134

### DIFF
--- a/src/isar/config/settings.py
+++ b/src/isar/config/settings.py
@@ -210,3 +210,24 @@ class Settings(BaseSettings):
 
 
 settings = Settings()
+
+
+class RobotSettings(BaseSettings):
+    def __init__(self) -> None:
+        try:
+            with pkg_resources.path(
+                f"{settings.ROBOT_PACKAGE}.config", f"settings.env"
+            ) as path:
+                env_file_path = path
+        except ModuleNotFoundError:
+            env_file_path = None
+        super().__init__(_env_file=env_file_path)
+
+    CAPABILITIES: List[str] = Field(default=["drive_to_pose", "take_image"])
+
+    class Config:
+        env_file_encoding = "utf-8"
+        case_sensitive = True
+
+
+robot_settings = RobotSettings()
--- a/src/isar/mission_planner/mission_validator.py
+++ b/src/isar/mission_planner/mission_validator.py
@@ -1,0 +1,9 @@
+from typing import List
+
+from isar.models.mission import Mission
+
+
+def is_robot_capable_of_mission(
+    mission: Mission, robot_capabilities: List[str]
+) -> bool:
+    return all([task.type in robot_capabilities for task in mission.tasks])
--- a/src/isar/apis/schedule/start_mission.py
+++ b/src/isar/apis/schedule/start_mission.py
@@ -1,15 +1,17 @@
 import logging
 from http import HTTPStatus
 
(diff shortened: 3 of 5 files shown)


src/isar/mission_planner/mission_validator.py
```diff
@@ -0,0 +1,9 @@
+from typing import List
+
+from isar.models.mission import Mission
+
+
+def validate_robot_capabilities(
+    mission: Mission, robot_capabilities: List[str]
+) -> bool:
+    return all([task.type in robot_capabilities for task in mission.tasks])
```
tsundvoll, line 9: Suggestion: Fetch the robot capabilities here instead
```suggestion
def is_robot_capable_of_mission(mission: Mission) -> bool:
    robot_capabilities = robot_settings.CAPABILITIES
    return all([task.type in robot_capabilities for task in mission.tasks])
```

vetlek:
[Linked PR in isar-robot](https://github.com/equinor/isar-robot/pull/31)

vetlek:
[Linked PR in isar-turtlebot](https://github.com/equinor/isar-turtlebot/pull/94)

src/isar/config/settings.py
```diff
@@ -210,3 +210,28 @@ class Config:
...
+        super().__init__(_env_file=env_file_path)
+
+    CAPABILITIES: List[str] = Field(default_factory=default_capabilities)
+
+    class Config:
+        env_file_encoding = "utf-8"
+        case_sensitive = True
+
+
+robot_settings = RobotSettings()
```
tsundvoll, line 237: I suggest to make this a 'private' member function to keep it gathered together where it is used.
```suggestion

class RobotSettings(BaseSettings):
    def __init__(self) -> None:
        try:
            with pkg_resources.path(
                f"{settings.ROBOT_PACKAGE}.config", f"settings.env"
            ) as path:
                env_file_path = path
        except ModuleNotFoundError:
            env_file_path = None
        super().__init__(_env_file=env_file_path)

    def _default_capabilities():
        return ["drive_to_pose", "take_image"]

    CAPABILITIES: List[str] = Field(default_factory=_default_capabilities)

    class Config:
        env_file_encoding = "utf-8"
        case_sensitive = True


robot_settings = RobotSettings()
```

tsundvoll (approved):
I suggest to move the factory inside the class. When that is fixed, I approve of this PR.

src/isar/config/settings.py
```diff
@@ -210,3 +210,24 @@ class Config:
...
+        try:
+            with pkg_resources.path(
+                f"{settings.ROBOT_PACKAGE}.config", f"settings.env"
+            ) as path:
+                env_file_path = path
+        except ModuleNotFoundError:
+            env_file_path = None
+        super().__init__(_env_file=env_file_path)
+
+    CAPABILITIES: List[str] = Field(default=["drive_to_pose", "take_image"])
```
tsundvoll, line 226: I would say this looks even better :)

PR #243, opened 2022-04-19
Add RobotInvalidTaskException

knord94:
Exception for invalid tasks which should not result in aborting mission but skipping the task.


vetlek (approved):
LGTM

PR #246, opened 2022-04-19
Add flake8 to project

tsundvoll:


anetteu (approved):
Looks good :+1:

setup.cfg
```diff
@@ -11,6 +11,16 @@ no_strict_optional = True
 no_site_packages = True
 ignore_missing_imports = True
 
+[flake8]
+filename =
+    ./src/*,
+    ./tests/*,
+exclude = .git, __pycache__, *.egg-info, *.json, *.conf, *.env,
+per-file-ignores =
+    __init__.py: F401,
```
Christdej, line 20: We could consider using
`__all__`
in the `__init_.py` - files where this is happening. Thoughts?

tsundvoll: Hm.. What if there is for instance a very long line in the `__init_.py` files? Can't come up with other examples, but maybe we should ignore specific checks for now, and if it becomes a problem with other checks, ignore them when it happens?

Christdej: Agreed, LGTM

PR #247, opened 2022-04-20
Move geometry to alitra

Christdej:
Depends on new version of Alitra https://github.com/equinor/alitra/pull/44


Christdej:
FYI, tests will fail until a new version of Alitra is deployed

src/isar/apis/schedule/drive_to.py
```diff
@@ -49,12 +46,12 @@ def post(
         if not ready:
             message, status_code = response
             return JSONResponse(content=asdict(message), status_code=status_code)
-
-        position: Position = Position(x=x, y=y, z=z, frame=Frame.Robot)
+        robot_frame = Frame("robot")
```
aestene, line 49:
```suggestion
        robot_frame: Frame = Frame("robot")
```

src/isar/mission_planner/echo_planner.py
```diff
@@ -126,11 +124,8 @@ def _get_robot_pose(self, tag_name: str) -> Pose:
         solution.
         """
         predefined_pose: Pose = predefined_poses[tag_name]
-
-        if predefined_pose.frame is Frame.Robot:
-            raise ValueError(
-                f"Frame of predefined pose should be {Frame.Asset} not {Frame.Robot}"
-            )
+        if predefined_pose.frame is Frame("robot"):
+            raise ValueError("Frame of predefined pose should be asset not robot")
```
aestene, line 128: Are you certain this will actually work?

Using the `is` operator in python between two classes only checks that the left and right operands have the same type as far as I know. I don't think it checks if the value of `name` within the class itself is the same. 

If you used `==` instead I think it also wouldn't work because they would be two different objects.

An alternative solution could be to override the `__eq__` method for the `Frame` class and say that two frames with the same name are equal.

Christdej: Good catch, didnt notice it was a `is` operator. `__eq__` will work as dataclasses implement their own `__eq__` that checks that the fields are the same between dataclasses

src/isar/apis/schedule/drive_to.py
Christdej: Fixed

aestene (approved):
LGTM. Await merging for tests to pass following new version of Alitra.

PR #249, opened 2022-04-20
Bug - Printing wrong config & Logging info when warning

GodVenn:
The printed config did not show which robot package was being used, and also did not include any module names, leaving the user guessing what the default value was.
The logger logged failed connect attempts as info when it should be warning

### DIFF
--- a/src/isar/modules.py
+++ b/src/isar/modules.py
@@ -1,15 +1,16 @@
+import logging
+from logging import Logger
 from importlib import import_module
 from types import ModuleType
-from typing import Any, List, Tuple
+from typing import List, Union
 
-from injector import Module, multiprovider, provider, singleton
+from injector import Injector, Module, multiprovider, provider, singleton
 
 from isar.apis.api import API
 from isar.apis.schedule.drive_to import DriveTo
 from isar.apis.schedule.start_mission import StartMission
 from isar.apis.schedule.stop_mission import StopMission
 from isar.apis.security.authentication import Authenticator
-from isar.config.configuration_error import ConfigurationError
 from isar.config.keyvault.keyvault_service import Keyvault
 from isar.config.settings import settings
 from isar.mission_planner.echo_planner import EchoPlanner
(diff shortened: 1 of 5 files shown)


src/isar/modules.py
```diff
@@ -177,26 +178,33 @@ def provide_mqtt_client(self, queues: Queues) -> MqttClientInterface:
         return None
 
 
+injector_modules: List[Module] = []
+module_config_keys: List[str] = []
```
aestene, line 182: What is the reasoning behind moving these variables outside the function itself? This way they will always be in memory as "global" variables in this file while I'm not certain they need to be?

aestene (changes requested):
Have a look at the one comment.

Otherwise everything looks good, good improvements!

src/isar/modules.py
GodVenn: It was so that the _append_key function would have access to them as if they where all inside a static class. 
The alternative is to pass them to each call to that function.
I'm not that comfortable with python yet, so if you think that's a better alternative then let's do that!

aestene: @Christdej @tsundvoll 

Any thoughts on this? Personally something feels off to me about having them as open variables within the file in this manner as they are then exported and it's possible to import them. But I'm not certain if it's a bad or good pattern.

tsundvoll: I agree that it looks a bit off. It can be a symptom of the entire file being written a bit hard to read. I am looking into a better way to formulate this now.

GodVenn: Maybe moving the functionality of this last part of the file to its own file could be a solution?

aestene: It could be moved but then again it is very tightly coupled with the modules defined in this file.

I think I'll say approved from my side, but let @tsundvoll comment before you merge to see if he thought of a smooth solution.

aestene (approved):
LGTM

Christdej (changes requested):
Me and Thomas discussed this yday, dont merge yet

src/isar/modules.py
tsundvoll: We found a smooth solution that resolves this comment.

Christdej (approved):

PR #254, opened 2022-04-25
Rename task to step

aestene:


anetteu:
This is a breaking change and should probably trigger a new release?

PR #256, opened 2022-04-27
Add get pose to interface

tsundvoll:
This requires implementation in all robots implementing the `RobotInterface`,
e.g. https://github.com/equinor/isar-turtlebot/pull/109

This PR will reimplement the MQTT module and call it `Telemetry`

Important change to notice is that the `StateMachine` now has one method, `publish_telemetry_thread()` that regularly publishes relevant telemetry payloads.


src/isar/config/settings.py
```diff
@@ -89,11 +89,10 @@ class Settings(BaseSettings):
     # Client ID for the OpenAPI client
     OPENAPI_CLIENT_ID: str = Field(default="5f412c20-8c36-4c69-898f-d2b5051f5fb6")
 
-    # MQTT username
+    # MQTT credentials
     # The username and password is set by the MQTT broker and must be known in advance
```
GodVenn, line 93: Isn't the username and password *used by* the MQTT broker? And *set by* the user?

src/isar/state_machine/state_machine.py
```diff
@@ -321,8 +286,39 @@ def _check_dependencies(self):
                 return False
         return True
 
+    def publish_telemetry_thread(self) -> None:
+        robot_id = settings.ROBOT_ID
+        PUBLISH_FREQUENCY = 1
```
GodVenn, line 291: Should this be in the settings?

src/isar/state_machine/state_machine.py
```diff
@@ -321,8 +286,39 @@ def _check_dependencies(self):
...
+
+        messenger = Telemetry()
+
+        while True:
+            timestamp = datetime.now().strftime("%Y-%m-%dT%H:%M:%SZ")
+
+            state: IsarStatePayload.State = IsarStatePayload.State(self.current_state)
+            payload_state = IsarStatePayload(state=state, timestamp=timestamp).to_json()
```
GodVenn, line 299: Could the conversion from `self.current_state` to `state` be done in the constructor of `IsarStatePayload` instead?

src/isar/telemetry/telemetry.py
```diff
@@ -0,0 +1,87 @@
...
+            "Failed to connect to MQTT Broker within set backoff strategy."
+        )
+
+    @backoff.on_exception(
+        backoff.expo,
+        ConnectionRefusedError,
+        max_time=300,
+        on_success=_on_success,
+        on_backoff=_on_backoff,
+        on_giveup=_on_giveup,
```
GodVenn, line 82: Love this!

GodVenn (commented):
Really nice and clean!
Please see the few comments:)

src/robot_interface/models/telemetry/entity.py
```diff
@@ -0,0 +1,14 @@
...
+    """This is the base class for the model classes. It provides json serialization
+    methods."""
+
+    @classmethod
+    def from_json(cls, data):
+        jsonObj = json.loads(data)
+        return cls(**jsonObj)
+
+    def to_json(self):
+        return json.dumps(self, default=lambda o: o.__dict__, sort_keys=True, indent=2)
```
aestene, line 14: Where are these classes from? Weren't they supposed to be generated and imported from Flotilla?

src/isar/state_machine/state_machine.py
```diff
@@ -321,8 +286,39 @@ def _check_dependencies(self):
...
+                z=pose.position.z,
+                yaw=euler_orientation[0],
+                pitch=euler_orientation[1],
+                roll=euler_orientation[2],
+            )
+            payload_pose = TelemetryPosePayload(
+                pose=pose_payload, timestamp=timestamp
+            ).to_json()
+
+            messenger.publish(f"isar/{robot_id}/state", payload_state)
```
aestene, line 320: Should the state be published at a second interval? It's a timeseries that only changes once in a while and it would be better to publish on change rather than continuously.

src/isar/state_machine/state_machine.py
```diff
@@ -321,8 +286,39 @@ def _check_dependencies(self):
...
+
+            state: IsarStatePayload.State = IsarStatePayload.State(self.current_state)
+            payload_state = IsarStatePayload(state=state, timestamp=timestamp).to_json()
+
+            battery_payload: int = 100
+            payload_battery = TelemetryBatteryPayload(
+                battery=battery_payload, timestamp=timestamp
+            ).to_json()
+
+            pose: Pose = self.robot.get_pose()
```
aestene, line 306: Conceptual question. Should ISAR publish pose and battery status? These are timeseries coming directly from the robot and some robots may provide their own MQTT streaming service which we can connect the broker to. Does the functionality of doing this belong in ISAR or `isar_robot`?

src/isar/state_machine/state_machine.py
```diff
@@ -321,8 +286,39 @@ def _check_dependencies(self):
...
+            euler_orientation: np.ndarray = pose.orientation.to_euler_array()
```
aestene, line 307: Why do we prefer to publish in euler rather than keep the quaternions?

src/robot_interface/robot_interface.py
```diff
@@ -90,3 +92,37 @@ def get_inspections(self, step: InspectionStep) -> Sequence[Inspection]:
...
+        Pose
+            An Alitra pose containing a position, an orientation and a frame.
+
+        Raises
+        ------
+        RobotException
+            If the pose can't be retrieved.
+
+        """
+        raise NotImplementedError
```
aestene, line 111: Does this scale?

Pose and battery status is not the only timeseries available from the robots and there will be different timeseries for different robots. Would we need to make individual interface functions for each type of data we wish to stream?

setup.cfg
```diff
@@ -10,6 +10,7 @@ follow_imports = normal
 no_strict_optional = True
 no_site_packages = True
 ignore_missing_imports = True
+exclude = build
```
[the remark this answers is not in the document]
tsundvoll: I suddenly got a build folder that caused mypy to complain (something with duplicated files). I deleted the folder and see that it has not reappeared. I will remove this line for now and if it becomes a problem again I will add it in.

src/isar/apis/schedule/start_mission.py
```diff
@@ -65,6 +65,10 @@ def post(
         )
         if not robot_capable:
             response.status_code = HTTPStatus.BAD_REQUEST
+            self.logger.error(
+                "Robot don't have necessary capabilities for the given mission"
```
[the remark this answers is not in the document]
tsundvoll: Well observed, will fix

src/isar/config/settings.py
tsundvoll: Absolutely, correcting to _used_

tsundvoll: I think the reason for the wording is that the MQTT broker is another "agent" than this ISAR package, and it is in the MQTT broker that the user actually sets the password. Still, we avoid the confusion by writing _used_ here. Well observed.

tsundvoll:
```python
    # The username and password for connecting to the MQTT broker
```

src/isar/state_machine/state_machine.py
tsundvoll: The `PUBLISH_FREQUENCY` you mean? That makes sense. I can put it there.

src/isar/state_machine/state_machine.py
tsundvoll: Yes, it could be done, and agree that would look much better. The reason I don't want to change `IsarStatePayload` for now is that we probably will fetch these models directly from `flotilla-openapi` where they might be auto-generated. I will have this in mind though.

src/isar/telemetry/telemetry.py
tsundvoll: Thanks! I can not take the credit for those lines though, this file is mostly a renaming and clean-up of the previous mqtt_client.py

src/robot_interface/models/telemetry/entity.py
tsundvoll: Yes, they are here just to prove the concept. They will be imported from flotilla-openapi.

src/isar/state_machine/state_machine.py
tsundvoll: The reason was the draft form mqtt-api in flotialla-openapi, but I agree it makes sense to continue with quaternions

src/isar/state_machine/state_machine.py
aestene: Another reason for publishing on changes is that then we can treat it like an event. If we publish every second Flotilla will have to make queries to the database every second for other topics that will show the current task status. Thus we need to support event based topics in some way which isn't possible with the current thread framework.

src/isar/state_machine/state_machine.py
```diff
@@ -264,33 +256,6 @@ def stop_mission(self):
...
-    def publish_mission(self) -> None:
-        payload: str = json.dumps(
-            {"mission": self.current_mission}, cls=EnhancedJSONEncoder
-        )
-
-        self.mqtt_client.publish(
-            topic=settings.ISAR_MISSION,
-            payload=payload,
-            retain=True,
-        )
```
aestene, line 292: These have been removed but as far as I can tell not replaced. We do want to change them a little so we're not publishing the full mission every time etc. but we do need to publish these on change so the functionality should be there.

See my comment below on possibly needing event based publishing.

src/isar/telemetry/telemetry.py
```diff
@@ -0,0 +1,87 @@
...
+
+from isar.config.settings import settings
+from robot_interface.models.telemetry.entity import Entity
+
+
+class Telemetry:
+    """This is a wrapper for the mqtt client."""
+
+    def __init__(self) -> None:
+        self.logger = logging.getLogger("mqtt_client")
```
aestene, line 15:
```suggestion
        self.logger = logging.getLogger("telemetry")
```
Rename all places the `mqtt_client` logger is used to `telemetry`. It's a better name.

src/isar/state_machine/state_machine.py
```diff
@@ -321,8 +286,38 @@ def _check_dependencies(self):
                 return False
         return True
 
+    def publish_telemetry_thread(self) -> None:
+        robot_id = settings.ROBOT_ID
+
+        messenger = Telemetry()
+
+        while True:
+            timestamp = datetime.now().strftime("%Y-%m-%dT%H:%M:%SZ")
```
aestene, line 295: Consider creating timestamp after each API call

src/isar/state_machine/state_machine.py
vetlek: I think that this functionality should  be done in `isar_robot`, but initiated from ISAR. The robot interface will grow and require new releases when we want to publish some new information to the broker and it might be that some robots havn't got the same functionalities. Could have that the robots need to implement a publish telemetry function that takes in an mqtt broker from ISAR with some config and is started in its own thread.

src/isar/state_machine/state_machine.py
vetlek: We should definitely treat it as an event and not like time series data.

src/robot_interface/robot_interface.py
vetlek: Agree

vetlek (changes requested):

aestene:

aestene:
Closing as this will be adressed by #289

PR #257, opened 2022-05-02
Add task level

aestene:


src/isar/models/mission.py
```diff
@@ -1,37 +1,60 @@
...
+    status: StepStatus = field(default=StepStatus.NotStarted, init=False)
+    id: UUID = field(default_factory=uuid4, init=False)
+    iterator: Iterator = None
+
+    def next_step(self) -> Step:
+        return self.iterator.__next__()
+
+    def is_finished(self) -> bool:
+        for step in self.steps:
+            if step.status is StepStatus.Failed or step.status is StepStatus.Unexpected:
```
anetteu, line 23: Remove `or step.status is StepStatus.Unexpected:` ? This case is covered by the elif below.

src/isar/models/mission.py
```diff
@@ -1,37 +1,60 @@
...
+                # One step has failed meaning the task as a whole should be aborted
+                self.status = StepStatus.Failed
```
anetteu, line 25: With the current implementation, the task will always fail if a step of the task fails and the whole task cannot be completed. Is this the desired behavior, or do we want the task to continue? In the case of a task with multiple inspection steps, the next inspection step might succeed even if the first inspection step fails.

src/isar/models/mission.py
```diff
@@ -1,43 +1,66 @@
...
```
tsundvoll, line 25: There is a difference between types of steps here. If a DriveToPose step fails, I agree that the task as a whole should fail. But if a task consisting of [DriveToPose, TakeImage and TakeThermalImage] and the TakeThermalImage fails, the TakeImage could still succeed and the task could be partly successful

src/isar/models/mission.py
```diff
@@ -1,43 +1,66 @@
...
-from robot_interface.models.mission import STEPS
-from robot_interface.models.mission.step import DriveToPose, InspectionStep
+from robot_interface.models.mission import STEPS, Step, StepStatus
 
 
 @dataclass
-class Mission:
+class Task:
     steps: List[STEPS]
+    status: StepStatus = field(default=StepStatus.NotStarted, init=False)
```
anetteu, line 14: Does it makes sense that `Task.status` have the type `StepStatus`?

anetteu (changes requested):

src/isar/models/mission.py
aestene: That is a good point. What is the best way to deal with it though?

Should a task conceptually be defined as a "requires all to pass for task to pass"? Because that is much easier to define and work with in practice. This would require that we split the `TakeImage` and `TakeThermalImage` in two tasks in your example to circumvent the issue though.

src/isar/models/mission.py
aestene: Thank you!

src/isar/models/mission.py
aestene: I was thinking of renaming it to simply `Status` or something? They should be the same for both `Task` and `Step` I think. I didn't want to create a new `TaskStatus` enum with the same entries.

src/isar/models/mission.py
aestene: I agree, same discussion as with @tsundvoll above. Conceptually I think it's easier to say that a `Task` should require all steps to succeed, otherwise it will be a nightmare to infere the status of the task based on so many conditions. However, this would make it more natural to have multiple tasks for multiple inspections of the same tag, which again moved the mission away from the input list mission.

src/isar/models/mission.py
tsundvoll: Yes, having a `required` attribute on the required steps could be a good solution.

Not sure if this is possible to set on the type of task, i.e. is a DriveToPose always required, and is a TakeImage always optional?

aestene: Then again, if you had only a `DriveToPose` and `TakeImage` you would say the `Task` failed if the `TakeImage` step which isn't listed as required failed.

I would say a `Task` has failed even if one out of X inspections have failed.

We could implement something along the line of

```
if MotionStep has failed:
    task has failed
elif InspectionStep has failed and steps remaining:
    task has partly failed but still perform remaining steps
```

And some more logic probably.

src/isar/mission_planner/echo_planner.py
```diff
@@ -50,7 +50,7 @@ def get_mission(self, mission_id: int) -> Mission:
             self.logger.error(msg)
             raise MissionPlannerError(msg) from e
 
-        mission: Mission = Mission(steps=[])
+        mission: Mission = Mission(tasks=[])
```
tsundvoll, line 53: Suggestion: Start with an empty list `tasks=[]` here, and then instantiate the `Mission` after the `tasks` list is populated

src/isar/mission_planner/echo_planner.py
```diff
@@ -75,14 +75,12 @@ def get_mission(self, mission_id: int) -> Mission:
...
                 )
                 continue
+            task: Task = Task(steps=[drive_step, *inspection_steps])
+            mission.tasks.append(task)
 
-            mission.steps.append(drive_step)
-            mission.steps.extend(inspection_steps)
-
-        if not mission.steps:
+        if not mission.tasks:
```
tsundvoll, line 81: Check length of `tasks` here and if > 0, instantiate mission

src/isar/models/mission.py
anetteu: Based on the discussion, I see three options: 

1. Leave the implementation as it is, requiring all steps to succeed for a  `Task` to succeed. The drawback with this solution is that if the task have several inspection steps, some of them might have been able to succeed.

2. Implement one `Task` per `InspectionStep`.

3. Let the `Task` fail if a `MotionStep` fails. This allows all `InspectionsStep`s to be executed. However, what will then be the requirement for the `Task` to succeed? At least one successful `InspectionStep`?

What are your thoughts, @aestene @tsundvoll?

tsundvoll: This is related to dependencies between steps, right?
If this is the case, I think a task should only fail if it does not make sense to continue the task. (Closest to option 3 from  @anetteu ). Don't abort the task if a measure step fails, but abort the task if a drivetopose step fails

tsundvoll: A PartialSuccessful status could be used to indicate partially successful tasks where one or more measure steps have failed.

aestene: There is also a difference between aborting the task and considering the task as failed. I would say nr. 3 with information to the user that the task has only partly succeeded. Should we add a new possible status for a task? `Status.PartlySuccessful`

src/isar/mission_planner/echo_planner.py
aestene: `if not list` is the pythonic way of checking if the list is empty.

But the instantiation of the mission may be moved down.

tsundvoll: Sure, that is better

src/isar/models/mission.py
```diff
@@ -47,3 +70,6 @@ def __post_init__(self) -> None:
 
         if self.metadata is None:
             self.metadata = MissionMetadata(mission_id=self.id)
+
+        if self.iterator is None:
+            self.iterator = iter(self.tasks)
```
tsundvoll, line 75: This looks a bit strange here. I assume it is done like this because we previously set tasks on the mission after the mission was instantiated. But now that this is changed (see other comment), I think we can move it to a post_init as it is done in `Task`

src/isar/models/mission.py
tsundvoll: A common status enum sounds good. I suggest the following. This should be sufficient, right?
```
class Status(str, Enum):
    NotStarted: str = "not_started"
    InProgress: str = "in_progress"
    Successful: str = "successful"
    PartlySuccessful: str = "partly_successful"
    Failed: str = "failed"
```

src/isar/config/predefined_missions/default_turtlebot.json
```diff
@@ -1,89 +1,104 @@
...
         },
         {
-            "type": "take_image",
-            "target": {
-                "x": 5.6,
-                "y": 5.2,
-                "z": 0,
-                "frame": "robot"
-            }
+            "steps": []
```
tsundvoll, line 69: Remove empty task?

src/isar/mission_planner/echo_planner.py
aestene: Done

src/isar/mission_planner/echo_planner.py
aestene: Done

src/isar/models/mission.py
aestene: It is already in the `__post_init__` function I believe. I'll rename it to also be internal.

src/isar/models/mission.py
```diff
@@ -1,43 +1,66 @@
...
+                continue
+            else:
+                return False
+
+        self.status = StepStatus.Completed
+        return True
+
+    def __post_init__(self) -> None:
+        if self.iterator is None:
+            self.iterator = iter(self.steps)
```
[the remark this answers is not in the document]
aestene: Done

src/isar/config/predefined_missions/default_turtlebot.json
aestene: Nicely spotted, done.

src/isar/models/mission.py
aestene: See conversation above.

src/isar/models/mission.py
aestene: There is also another enum named `Status` in ISAR which indicated the status of the state machine. I will see what makes sense as a change.

aestene: And I prefer `partial_success` tbh. Are we removing `Scheduled` at the same time?

tsundvoll: Do we need scheduled? Is it not the same as not started?

tsundvoll: > There is also another enum named `Status` in ISAR which indicated the status of the state machine. I will see what makes sense as a change.

This could maybe just be attributes of the `StateMachine` object?

aestene: And removed scheduled. The one place it was used I set it to `InProgress` instead as it made more sense.

src/isar/state_machine/state_machine.py
```diff
@@ -298,28 +301,19 @@ def _log_state_transition(self, next_state):
...
-                return False
-        return True
+        log_statements: List[str] = []
+        for task in mission.tasks:
+            log_statements.append(
+                f"{type(task).__name__:<20} {str(task.id)[:8]:<32} -- {task.status}"
+            )
+            for j, step in enumerate(task.steps):
+                log_statements.append(
+                    f"{j:>3} {type(step).__name__:<20} {str(step.id)[:8]:<32} -- {step.status}"
```
tsundvoll, line 311: Too long line, but it may make sense to keep it thus, due to its complexity. Add the `noqa` tag to ignore warnings from `flake8`
```suggestion
                    f"{j:>3} {type(step).__name__:<20} {str(step.id)[:8]:<32} -- {step.status}"  # noqa: E501
```

src/isar/state_machine/state_machine.py
```diff
@@ -80,9 +79,9 @@ def __init__(
         self.sleep_time = sleep_time
 
         self.mission_in_progress: bool = False
-        self.current_mission: Mission = Mission(steps=[])
+        self.current_mission: Mission = Mission(tasks=[])
```
tsundvoll, line 82: Should not `self.current_mission` be initialized to `None` before `start_mission()` is called (which populates the `current_mission`, and also be set to `None` again when the mission is done in `reset_state_machine()`?

tsundvoll (approved):
A couple of more linting suggestions, but I approve of the content.

I think we still can re-structure the flow of things to improve even further, but this feature of adding a task level to the hierarchy in itself looks good to me, well done :+1:

src/isar/models/mission.py
aestene: This has now been implemented in such a way that a task will be `PartiallySuccessful` if one or more `InspectionSteps` have failed. If all `InspectionSteps` have failed the entire task will be considered as failed.

If a `MotionStep` fails the entire task is considered as failed. There is an underlying assumption here that most tasks will only contain one `MotionStep` which I think is fair.

aestene: When/if we wish to deal with multiple inspections of the same target from different angles we'll look into this some more.

src/isar/models/mission.py
```diff
@@ -1,43 +1,96 @@
...
+            if step.status is StepStatus.Failed and isinstance(step, MotionStep):
+                # One motion step has failed meaning the task as a whole should be aborted
+                self.status = StepStatus.Failed
+                return True
+
+            elif (step.status is StepStatus.Failed) and isinstance(
+                step, InspectionStep
+            ):
+                # It should be possible to perform several inspections per task. If one out of many inspections fail
+                # the task is considered as partially successful.
```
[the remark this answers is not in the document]
aestene: Thank you, my PyCharm line separator was set to 120 so I didn't notice.

anetteu (approved):

anetteu:
I Agree with @tsundvoll, good work! Please rewrite the commit message to reflect the changes made after our discussion before merging

src/isar/state_machine/states/monitor.py
```diff
@@ -95,7 +96,12 @@ def _queue_inspections_for_upload(self, current_step: InspectionStep):
             )
             return
 
-        mission_metadata: MissionMetadata = self.state_machine.current_mission.metadata
+        # As this object is passed to another thread it should be frozen at this
+        # point a deepcopy is used
```
tsundvoll, line 100: Suggestion:
```suggestion
        # A deepcopy is made to freeze the metadata before passing it to
        # another thread through the queue
```

tsundvoll (approved):

PR #259, opened 2022-05-10
Add video steps and inspections

### LINKED ISSUE #258: Add video (visual and thermal) to the list of supported steps
Steps for initiating recording of video (both thermal and visual) should be added to the list of supported steps. This should also include adding these steps based on the mission received from Echo. The user should be able to specify the duration of the video.

A step in the list of steps in the robot_interface.  

No change.

vetlek:
Closes #258


aestene (approved):
LGTM

PR #268, opened 2022-05-18
Update api responses

### LINKED ISSUE #265: Update start response to include an overview of the mission started
When successfully starting a mission, the start endpoint should return a response containing an overview of the started mission with the id of the mission, list of all tasks (and steps) with the id, tag and inspection type. 

This will change the response returned from the endpoint. This will make it possible for consumers (like flotilla) to get references to the different tasks/steps before their status is published over MQTT.

No changes.

vetlek:
Closes #265

### DIFF
--- a/src/isar/apis/models/models.py
+++ b/src/isar/apis/models/models.py
@@ -1,14 +1,28 @@
-from typing import Optional
+from typing import List, Optional, Union
+from uuid import UUID
 
 from pydantic import BaseModel
 
 
-class StartResponse(BaseModel):
+class StartFailedResponse(BaseModel):
     message: str
-    started: bool
-    mission_id: Optional[str] = None
 
 
-class StopResponse(BaseModel):
+class StepResponse(BaseModel):
+    id: UUID
+    type: str
+
+
+class TaskResponse(BaseModel):
+    id: UUID
+    tag_id: Optional[str]
+    steps: List[StepResponse]
+
+
+class StartMissionResponse(BaseModel):
+    id: Union[UUID, int, str, None]
+    tasks: List[TaskResponse]
+
+
+class StopFailedResponse(BaseModel):
     message: str
-    stopped: bool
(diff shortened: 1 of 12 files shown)


src/isar/apis/models/models.py
```diff
@@ -1,14 +1,28 @@
...
+
+
+class TaskResponse(BaseModel):
+    id: UUID
+    tag_id: Optional[str]
+    steps: List[StepResponse]
+
+
+class StartMissionResponse(BaseModel):
+    id: Union[UUID, int, str, None]
```
GodVenn, line 23: Why are there so many options for the mission id?

src/robot_interface/models/mission/step.py
```diff
@@ -54,6 +54,9 @@ def robot_class_to_pretty_string(obj: Step) -> str:
         class_name: str = type(self).__name__
         return class_name + robot_class_to_pretty_string(self)
 
+    def to_dict(self) -> dict:
+        return {"id": self.id, "type": self.__class__.__name__}
```
GodVenn, line 58: Should this dict also contain the status of the step?

GodVenn (commented):
Please see comments for discussion

src/isar/apis/models/models.py
vetlek: This only reflects the possibilities given in the definition of the mission id. Not sure why there is so many different options for the id of a mission. This would be a separate issue.

src/robot_interface/models/mission/step.py
vetlek: At the beginning of the mission all steps and tasks are set as `NotStarted`, so there is no need to send this. A possibility is to include tasks that we were unable to plan (not able to find predefined poses/coordinates from STID, unsupported inspection types).

GodVenn: So this is only meant to be called in the beginning of the program?
Might be worth renaming it to reflect that.

GodVenn (approved):
LGTM

PR #272, opened 2022-05-24
Pause mission

### LINKED ISSUE #269: Add functionality to pause and resume a mission
It should be possible to pause and subsequently resume a mission that is in progress. It should also be possible to cancel a mission that is currently paused. 

It could be that adding a new state in the state machine is required for this functionality.

N/A

vetlek:
Closes #269


src/isar/state_machine/state_machine.py
```diff
@@ -64,23 +65,110 @@ def __init__(
...
+                    "source": self.initiate_step_state,
+                    "dest": self.initiate_step_state,
+                    "before": self._step_infeasible,
+                },
+                {
+                    "trigger": "step_failed",
+                    "source": self.initiate_step_state,
+                    "dest": self.idle_state,
+                    "before": self._step_failed,
+                },
```
aestene, line 157: What is actually happening in this transition? It's not `step_failed`?

I assume it's meant that we failed to schedule anything on the robot the ten times and thus cancel the mission?

src/isar/state_machine/state_machine.py
```diff
@@ -92,6 +180,84 @@ def __init__(
         self.transitions_log_length = transitions_log_length
         self.transitions_list: Deque[States] = deque([], self.transitions_log_length)
 
+    def _step_initiated(self) -> None:
+        self.current_step.status = StepStatus.InProgress
+        self.publish_step_status()
```
aestene, line 187: Should there be an
```
if self.mqtt_client:
```

src/isar/state_machine/state_machine.py
```diff
@@ -92,6 +180,84 @@ def __init__(
...
+
+    def _unpause(self) -> None:
+        self.current_mission.status = MissionStatus.InProgress
+        self.current_task.status = TaskStatus.InProgress
+        if self.mqtt_client:
+            self.publish_mission_status()
+            self.publish_task_status()
+
+    def _finalize(self) -> None:
+        self.publish_mission_status()
```
aestene, line 210: Should there be an
```
if self.mqtt_client:
```

src/isar/state_machine/state_machine.py
aestene: I see this occurs quite often so I wont comment everywhere

src/isar/state_machine/state_machine.py
```diff
@@ -92,6 +180,84 @@ def __init__(
...
+
+    def _pause(self) -> None:
+        self.paused = True
+
+    def _unpause(self) -> None:
+        self.current_mission.status = MissionStatus.InProgress
+        self.current_task.status = TaskStatus.InProgress
+        if self.mqtt_client:
+            self.publish_mission_status()
+            self.publish_task_status()
```
aestene: Having reviewed the code below it seems this if statement should no longer be required.

src/isar/state_machine/state_machine.py
```diff
@@ -64,23 +65,110 @@ def __init__(
...
+                    "trigger": "step_finished",
+                    "source": self.monitor_state,
+                    "dest": self.initiate_step_state,
+                    "before": self._step_finished,
+                },
+                {
+                    "trigger": "paused_successfully",
+                    "source": self.monitor_state,
+                    "dest": self.initiate_step_state,
+                    "before": self._paused_successfully,
```
aestene, line 144: Why is the destination state here not the paused state?

src/isar/state_machine/state_machine.py
aestene, line 210: Do we ever set mission status to `MissionStatus.Successful`?

src/isar/config/settings.env
```diff
@@ -1,4 +1,4 @@
-ISAR_ROBOT_PACKAGE = isar_robot
+ISAR_ROBOT_PACKAGE = isar_turtlebot
```
aestene, line 1: Revert to isar_robot

src/isar/state_machine/state_machine.py
```diff
@@ -92,30 +180,106 @@ def __init__(
...
+            f"Successfully initiated "
+            f"{type(self.current_step).__name__} "
+            f"step: {str(self.current_step.id)[:8]}"
+        )
+
+    def _pause(self) -> None:
+        self.paused = True
+
+    def _resume(self) -> None:
+        self.logger.info("Resuming mission!")
```
aestene, line 198: Log the ID of the mission that is being resumed

src/isar/state_machine/state_machine.py
```diff
@@ -92,30 +180,106 @@ def __init__(
...
+        self.paused = False
+        self.current_mission.status = MissionStatus.InProgress
+        self.current_task.status = TaskStatus.InProgress
+        self.publish_mission_status()
+        self.publish_task_status()
+        self.queues.resume_mission.output.put(True)
+        self.current_task.reset_task()
+        self.update_current_task()
```
aestene, line 206: Is this call necessary? It should just skip the entire function anyway

src/isar/state_machine/state_machine.py
```diff
@@ -92,30 +180,106 @@ def __init__(
...
+        self.logger.info(f"State transitions:\n  {state_transitions}")
+
+    def _mission_started(self) -> None:
+        self.logger.info(f"Starting new mission: {self.current_mission.id}")
+        self.current_mission.status = MissionStatus.InProgress
+        self.current_task.status = TaskStatus.InProgress
+        self.publish_mission_status()
+        self.publish_task_status()
+        self.log_step_overview(mission=self.current_mission)
+        self.update_current_task()
```
aestene, line 228: Is it necessary to call update_current_task? The task has already been set and the function does nothing.

src/isar/state_machine/state_machine.py
```diff
@@ -92,30 +180,106 @@ def __init__(
...
+        self.update_current_step()
+
+    def _step_finished(self) -> None:
+        self.publish_task_status()
```
aestene, line 232: Should the task status be published here? There is not necessarily a change in the task status and if there is a change it will be published in `update_current_task`

src/isar/state_machine/state_machine.py
vetlek: Guess `initiate_step_failed` would be a more descriptive name

src/isar/state_machine/state_machine.py
vetlek: Nope, we should do that and publish the updated status

src/isar/state_machine/state_machine.py
vetlek: Meant to be `self.publish_step_status()`

src/isar/state_machine/state_machine.py
vetlek: Agree

src/isar/state_machine/state_machine.py
vetlek: Agree

src/isar/state_machine/state_machine.py
```diff
@@ -92,6 +180,84 @@ def __init__(
...
+        self.logger.info(
+            f"Successfully initiated "
+            f"{type(self.current_step).__name__} "
+            f"step: {str(self.current_step.id)[:8]}"
+        )
+
+    def _pause(self) -> None:
+        self.paused = True
+
+    def _unpause(self) -> None:
```
oysand: Would a better word be continue?

src/isar/apis/schedule/pause_mission.py
```diff
@@ -0,0 +1,68 @@
...
+
+from isar.config.settings import settings
+from isar.models.communication.queues.queue_timeout_error import QueueTimeoutError
+from isar.models.communication.queues.queues import Queues
+from isar.services.utilities.queue_utilities import QueueUtilities
+from isar.services.utilities.scheduling_utilities import SchedulingUtilities
+from isar.state_machine.states_enum import States
+
+
+class PauseMission:
```
oysand, line 16: Could this be combined in a mission utilities file or class?

vetlek: Addressed in #274

src/isar/state_machine/states/stop.py
```diff
@@ -0,0 +1,59 @@
...
+            if self.state_machine.should_send_status():
+                self.state_machine.send_status()
+
+            try:
+                self.stop_thread.get_output()
+            except ThreadedRequestNotFinishedError:
+                time.sleep(self.state_machine.sleep_time)
+                continue
+            except RobotException:
+                continue
```
oysand, line 51: If we just continue this could "trap" the state machine if the robot refuse to stop for some reason. This could mean that the robot is not able to be stopped. In this case we would like to raise an alert. Should we consider some kind of retry amount before raising an alert?

src/isar/state_machine/states/stop.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def stop(self):
+        if self.stop_thread:
+            self.stop_thread.wait_for_thread()
+        self.stop_thread = None
+
+    def _run(self):
+        transition: Callable
+        while True:
+            if not self.stop_thread:
```
MortFred, line 38: check if we got a stop signal

src/isar/state_machine/states/stop.py
vetlek: Adding this as a separate issue since I am unsure what the desired behavior is for this case.

MortFred (approved):

PR #276, opened 2022-05-27
Combine api-controllers into single controller

### LINKED ISSUE #274: Combine all endpoint controllers into single class
Currently the controllers for the endpoints in the API is split into multiple files and classes. The different classes uses the same resources and are all related to controlling missions, thus they should be the same class. Suggest combining these into one single class called `MissionController` or similar.

No changes, but a more natural structure.

No changes

vetlek:
Closes #274


src/isar/apis/schedule/scheduling_controller.py
```diff
@@ -0,0 +1,208 @@
...
+        self.logger.info("Received request to stop current mission")
+
+        try:
+            state: Optional[States] = self.scheduling_utilities.get_state()
+        except:
+            self.logger.error("Timeout - Could not get current state of state machine")
+            response.status_code = HTTPStatus.REQUEST_TIMEOUT.value
+            return
+
+        if state == States.Idle:
```
oysand, line 145: List all states it cannot be

src/isar/services/utilities/scheduling_utilities.py
```diff
@@ -25,27 +25,28 @@ def __init__(self, queues: Queues, queue_timeout: int = settings.QUEUE_TIMEOUT):
...
-            return QueueUtilities.check_queue(
-                self.queues.mission_status.output, self.queue_timeout
-            )
-        except QueueTimeoutError:
-            self.logger.error(
-                "Timeout while trying to get the state of the state machine"
-            )
-            QueueUtilities.clear_queue(self.queues.mission_status.input)
-            return None
+        return self._send_command(True, self.queues.mission_status)
```
oysand, line 28: Rename mission status queue to get state

oysand (approved):
LGTM

PR #278, opened 2022-05-30
Initialization of robot in state machine

### LINKED ISSUE #273: Add initialization of robot
Robots might need different types of preparations (localization etc) before they are ready to perform missions. These types of actions should be done before the robot starts a mission/is able to start a mission. 

A suggested solution is to add a `def prepare(self, prepare_params)` to the robot interface. A new state is created (either before entering idle or between idle and initiate_step) that make use of this function and depending on the result either aborts (notifies users that the robot is not ready) or continue to the next state. 

No changes

vetlek:
Closes #273


src/isar/apis/models/models.py
```diff
@@ -1,3 +1,4 @@
+from email.policy import default
```
GodVenn, line 1: Is this used for anything?

src/isar/state_machine/states/initialize.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class Initialize(State):
+    @inject
+    def __init__(self, state_machine: "StateMachine"):
+        super().__init__(name="initialize", on_enter=self.start, on_exit=self.stop)
+        self.state_machine: "StateMachine" = state_machine
+
+        self.logger = logging.getLogger("state_machine")
+        self.initialize_thread: Optional[ThreadedRequest] = None
```
GodVenn, line 25: If the state is already called `Initialize`, shouldn't the thread simply be named `thread` ?
`initialize_thread` sounds like a function name to me.

src/isar/state_machine/states/initialize.py
```diff
@@ -0,0 +1,60 @@
...
+                )
+
+            try:
+                self.initialize_thread.get_output()
+            except ThreadedRequestNotFinishedError:
+                time.sleep(self.state_machine.sleep_time)
+                continue
+            except RobotException as e:
+                self.logger.error("Initialization of robot failed")
+                self.logger.error(e)
```
GodVenn, line 54: Should this be one log call instead of two?

GodVenn (commented):
Great work!
Please see comments

src/isar/state_machine/states/initialize.py
vetlek: The name reflects the name of the function (see the other states for example), not the state.

GodVenn (approved):
LGTM

PR #280, opened 2022-05-30
Publish state of state machine to api using queue

### LINKED ISSUE #279: Publish state of state machine using queue instead of checking for requests
Currently the endpoint controllers check the current state of the state machine to determine if the action should be processed further. This is done by sending a request to the state machine using queues and waiting on a response on a different queue. In the state machine, we are required to continuously monitor the request queue and send the state when required. This should be unnecessary. Instead the state of the state machine can be "published" when there are changes to the state and the endpoint controller can check the current state without first sending a request.

Less work needed in the state machine and more responsive API.

No changes.

vetlek:
Closes #279

### DIFF
--- a/src/isar/models/communication/queues/status_queue.py
+++ b/src/isar/models/communication/queues/status_queue.py
@@ -1,0 +1,20 @@
+from collections import deque
+from queue import Empty, Queue
+from typing import Any
+
+
+class StatusQueue(Queue):
+    def __init__(self) -> None:
+        super().__init__()
+
+    def check(self) -> Any:
+        if not self._qsize():
+            raise Empty
+        with self.mutex:
+            l = list(self.queue)
+            return l.pop()
+
+    def update(self, item: Any):
+        with self.mutex:
+            self.queue = deque()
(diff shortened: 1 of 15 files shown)


src/isar/state_machine/state_machine.py
```diff
@@ -341,6 +332,9 @@ def should_resume_mission(self) -> bool:
         except queue.Empty:
             return False
 
+    def send_state(self):
```
oysand, line 335: Update name so it is clear that this provides the state as a status message

src/isar/models/communication/queues/status_queue.py
```diff
@@ -0,0 +1,21 @@
+from collections import deque
+from queue import Empty, Queue
+from typing import Any
+
+
+class StatusQueue(Queue):
+    def __init__(self) -> None:
+        super().__init__(maxsize=1)
```
oysand, line 8: Maxsize is not use, consider removing it?

oysand (approved):
LGTM

PR #281, opened 2022-05-31
Add return to specific pose at end of mission

### LINKED ISSUE #270: Add return to home / docking functionality
There should be an endpoint for a robot to return to its home location. Home location could be docking or just a specified position.

- Consider if the position should be specified in the endpoint or be a configured variable
- Run a mission on the robot with a `DockingProcedure`

N/A

vetlek:
Possibly Closes #270


GodVenn (commented):
LGTM, but fix failing test
@aestene Does this close #270 ?

aestene:
Let's say this solves the current need and then we might have to build upon it later @GodVenn.

GodVenn (approved):
LGTM

PR #289, opened 2022-06-20
Add telemetry publisher support

aestene:
Closes #256


src/robot_interface/robot_interface.py
```diff
@@ -112,3 +115,17 @@ def initialize(self, params: InitializeParams) -> None:
 
         """
         raise NotImplementedError
+
+    @abstractmethod
+    def publish_telemetry(self, queue: Queue) -> List[Thread]:
```
oysand, line 120: Consider renaming

oysand (approved):
LGTM, this should be merged only when the robot packages are ready

PR #296, opened 2022-07-05
Update README

tsundvoll:
To reflect that isar-robot can now be installed from PyPI
instead of installing it from the github repository.


GodVenn (approved):
LGTM

PR #297, opened 2022-07-08
Add modular task selector interface

### LINKED ISSUE #261: Modularize planning functionality for selecting next task
The next task in a mission is today always chosen as the next element in the list using an iterator. However, it would be a nice feature to be able to modularize this functionality behind an interface such that different types of planners could be implemented. This would allow us to have planners that select the next task on other criteria, for example distance to location, battery level etc.

Utilize the existing setup for the current implementation of the `MissionPlannerInterface` and the `StorageInterface` to implement a new interface. A suggestion would be to rename the old `MissionPlanner` to `MissionLoader` and use `Planner` in this setting instead as it's conceptually more correct.

The interface would implement one function something like this
```
def select_next_task(tasks: List[Task]) -> Task:
```
Where the interface is passed the list of tasks in the current mission and returns which task should be performed next.

N/A

sondreo:
Closes #261.

This PR implements a modular interface for task selection per mission. It also replaces the iterative task selection from the Mission class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,68 @@ You can create your own storage module by implementing the [storage interface](.
 and adding your storage module to the selection [here](./src/isar/modules.py). Note that you must add your module as an
 option in the dictionary.
 
+## Task selection
+
+The tasks of a mission are selected based on a task selector module, defined by the `TASK_SELECTOR` configuration variable. The default task selector is `sequential`. When using the default module, tasks are executed in sequential order defined by the current input mission.
+
+### Implement you own task selector module
+
+Custom task selector modules may be added by implementing additional versions of the [task selector interface](./src/isar/mission_planner/task_selector_interface.py).
+
+For every custom module, the interface function `next_task()` must be implemented. All interface implementations by default have access to the list of tasks in the current mission through the member `self.tasks`, however additional variables may be supplied by adding arguments to `next_task()`. To comply with the interface definition, the function should return the next task upon every call, and raise the `TaskSelectorStop` exception when all tasks in the current mission have been completed:
+
+```python
+class CustomTaskSelector(TaskSelectorInterface):
+    ...
+    def next_task(...) -> Task:
+        
+        # Add code here
+        ...
+
+        # Raise `TaskSelectorStop` when all tasks have been completed
+        ...
+```
+
+Optionally, the `initialize()` function may be extended by supplementing the parameter list or function body:
+
+```python
+class CustomTaskSelector(TaskSelectorInterface):
+    ...
+    def initialize(self, tasks: List[Task], ...) -> None:
+        super.initialize(tasks=tasks)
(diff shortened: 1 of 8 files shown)


src/isar/mission_planner/task_selector_interface.py
```diff
@@ -0,0 +1,16 @@
...
+
+class TaskSelectorInterface(metaclass=ABCMeta):
+    def __init__(self) -> None:
+        self.tasks: List[Task] = None
+
+    def initialize(self, tasks: List[Task]) -> None:
+        self.tasks = tasks
+
+    @abstractmethod
+    def next_task(self) -> Task:
```
vetlek, line 15: How do you indicate that all tasks are finished? Should it return None or raise an exception? I see that in the state machine we expect a StopIteration exception when there are no more tasks left. This works fine for the IterativeTaskSelector, but should probably be generalized with an exception that is not specific to iterators. Suggest adding some documentation on the expected behavior of the next_task method.

vetlek:
What is the plan with regards to do the overall planning? I guess in most cases the planning will be done once based on some algorithm and then the tasks are provided in that order based on the first planning. Is this planning going to be done in the initialize function? Should there be an option to replan the remaining tasks?

src/isar/mission_planner/task_selector_interface.py
sondreo: Good point. Will generalize the stopping condition for updating the current task in the state machine.

sondreo:
> What is the plan with regards to do the overall planning? I guess in most cases the planning will be done once based on some algorithm and then the tasks are provided in that order based on the first planning. Is this planning going to be done in the initialize function? Should there be an option to replan the remaining tasks?

The idea is for the planning to be done either once-off, or alternatively based on dynamic criteria in `next_task`. Since the task selector instance is initialized with the list of tasks for the current mission, I believe it should be possible to implement a task selector module which re-plans dynamically over the course of a mission.

MortFred (commented):
Rename Iterative_select to sequential_select.
Update readme to show how to implement new selector implementation.

README.md
```diff
@@ -237,6 +237,68 @@ You can create your own storage module by implementing the [storage interface](.
 and adding your storage module to the selection [here](./src/isar/modules.py). Note that you must add your module as an
 option in the dictionary.
 
+## Task selection
+
+The tasks of a mission are selected based on a task selector module, defined by the `TASK_SELECTOR` configuration variable. The default selector is `sequential`.
```
aestene, line 242:
```suggestion
The tasks of a mission are selected based on a task selector module, defined by the `TASK_SELECTOR` configuration variable. The default selector is `sequential` indicating that tasks are executed in the sequential order defined by the input mission.
```

MortFred (approved):

PR #305, opened 2022-08-05
Remove latch

oysand:


GodVenn:
Why? 🤔

aestene:
The retain causes the latest message to be read when connecting to the broker. Whenever Flotilla connects to the broker this then creates an issue where messages related to previous missions on unknown robots are read. This is mostly an issue whenever we're testing and constantly recreating the database. However we risk getting an old message saying the robot is in a state which is not true based on this.

aestene (approved):
LGTM

PR #307, opened 2022-08-12
Add new dockerfile for turtlebot

aestene:
This is required as the Radix build pipeline does not support
component specific build variables which means we can't separate robot
packages for the different components.


GodVenn (approved):
LGTM

PR #308, opened 2022-08-12
Add logic to break stop_step loop if after a certain number of tries

eivindsjovold:


src/isar/state_machine/states/stop_step.py
```diff
@@ -20,6 +20,9 @@ def __init__(self, state_machine: "StateMachine"):
         self.state_machine: "StateMachine" = state_machine
         self.logger = logging.getLogger("state_machine")
         self.stop_thread = None
+        self._count_number_retries = 0
+        self._has_failed_current = False
+        self.has_failed_list: List[bool] = []
```
oysand, line 25: Do we need for this list for anything other than the tests?

src/isar/state_machine/states/stop_step.py
```diff
@@ -43,17 +48,32 @@ def _run(self):
...
-                continue
+                if self._has_failed_current:
+                    transition = self.state_machine.mission_stopped
+                    self.logger.warning(
+                        "Could not communicate request: Reached limit for stop attemps. Cancelled mission and transitioned to idle."
+                    )
+                    break
+                else:
+                    self.logger.info("Attempting to stop mission.")
+                    self.handle_stop_fail(retry_limit=50)
```
oysand, line 59: Could this function return the _has_failed_current to be used straight in the if statement at line 51?

src/isar/state_machine/states/stop_step.py
eivindsjovold: Kun i test foreløpig.

oysand: Then we should try to adjust the tests so that it is not needed

src/isar/state_machine/states/stop_step.py
```diff
@@ -20,6 +20,9 @@ def __init__(self, state_machine: "StateMachine"):
         self.state_machine: "StateMachine" = state_machine
         self.logger = logging.getLogger("state_machine")
         self.stop_thread = None
+        self._count_number_retries = 0
+        self._has_failed_current = False
```
aestene, line 24:
```suggestion
        self._count_number_retries: int = 0
        self._has_failed_current: bool = False
```

src/isar/state_machine/states/stop_step.py
aestene, line 25:
```suggestion
        self.has_failed: List[bool] = []
```

src/isar/state_machine/states/stop_step.py
```diff
@@ -20,6 +19,8 @@ def __init__(self, state_machine: "StateMachine"):
         self.state_machine: "StateMachine" = state_machine
         self.logger = logging.getLogger("state_machine")
         self.stop_thread = None
+        self._count_number_retries: int = 0
+        self._has_failed_current: bool = False
```
oysand, line 23: This should be removed

src/isar/state_machine/states/stop_step.py
```diff
@@ -29,6 +30,8 @@ def stop(self):
         if self.stop_thread:
             self.stop_thread.wait_for_thread()
         self.stop_thread = None
+        self._count_number_retries = 0
+        self._has_failed_current = False
```
oysand, line 34: This should be removed

tests/isar/state_machine/test_state_machine.py
```diff
@@ -20,6 +23,9 @@
 from tests.mocks.pose import MockPose
 from tests.mocks.robot_interface import MockRobot
 from tests.mocks.step import MockStep
+from pytest_mock import MockerFixture
+
+LOGGER = logging.getLogger("state_machine")
```
oysand, line 28: Is this used for anything?

oysand (approved):
LGTM

PR #310, opened 2022-08-18
Add action to post repository dispatch event

aestene:
This action will post a repository dispatch event on every merge to main which triggers a multi-repository build.


GodVenn (approved):
LGTM, great work!

PR #316, opened 2022-08-24
Configure log-levels through env variables

### LINKED ISSUE #250: Allow selecting log level for all threads in config
It should be possible to set the log level for all different loggers through the configuration. For example it should be possible to set the `api` logger to `debug` and have `state_machine`, `mqtt_client` and `uploader` be at `warning`.

The loggers are defined here https://github.com/equinor/isar/blob/main/src/isar/config/log.py although it seems the `mqtt_client` has not been added. In order to dynamically change the log levels the config may have to be moved into code rather than a `.conf` file which we currently use.

Not affected.

phhuseboe:
Closes #250

### DIFF
--- a/src/isar/config/settings.py
+++ b/src/isar/config/settings.py
@@ -204,6 +204,17 @@ class Settings(BaseSettings):
 
     TOPIC_ISAR_STEP: str = Field(default="step")
 
+    LOG_LEVELS: dict = Field(
+        default={
+            "console": "DEBUG",
+            "api": "DEBUG",
+            "urllib3": "WARNING",
+            "uvicorn": "WARNING",
+            "state_machine": "DEBUG",
+            "uploader": "DEBUG",
+        }
+    )
+
     @validator(
         "TOPIC_ISAR_ROBOT",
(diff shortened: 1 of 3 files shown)


src/isar/config/settings.py
```diff
@@ -204,6 +204,26 @@ class Settings(BaseSettings):
...
+    STATE_MACHINE_LOG_LEVEL: str = Field(default="INFO")
+    UPLOADER_LOG_LEVEL: str = Field(default="INFO")
+
+    LOG_LEVELS: dict = Field(default={})
+
+    @validator("LOG_LEVELS", pre=True, always=True)
+    def set_log_levels(cls, v, values) -> dict:
+        return {
+            "console": values["API_LOG_LEVEL"],
+            "api": values["CONSOLE_LOG_LEVEL"],
```
GodVenn, line 220: I believe these two have been switched

GodVenn (commented):
Looks good!

eivindsjovold:
LGTM

eivindsjovold (approved):

PR #317, opened 2022-08-26
Fix pose for start mission with ID

aestene:
There was a bug where the correct objects weren't being used for the endpoint to start a mission with a given ID.
- Additionally some formatting improvements.


GodVenn (approved):
LGTM

PR #318, opened 2022-08-29
Workflow for publishing docker images to registry

eivindsjovold:


oysand:
Should we have the publishing of the image in the robot package repo instead of isar? That way isar-robot publish isar build with isar-robot etc. We can then also lock the version of isar to the specific robot repo

aestene:

aestene:
This has been solved in other pull requests.

PR #327, opened 2022-09-21
Add option to disable tls through MQTT

### LINKED ISSUE #323: ISAR MQTT without SSL/TLS
I run mosquitto/mqtt locally without SSL/TLS setup. Need to comment out these lines in `src/isar/services/service_connections/mqtt/mqtt_client.py` line 45:
```
        #cert_path = os.path.join(dirname, "../../../config/certs/ca-cert.pem")
        #self.client.tls_set(ca_certs=cert_path) 
```

Maybe TLS could be disabled with a setting?

Leave blank for maintainers to fill out if you are uncertain about this.

### LINKED ISSUE #324: Robot state MQTT messages not retained?
If I start my own program that interfaces with ISAR *after* ISAR has started, then the robot state might be idle, and then I never get a new `state` message over MQTT. Previously, these messages were "retained", so that you would get them again when you re-connect. For example, when I restart my program, I would like to know the current state.

I see that you made the state messages not-retained. Why is that? Is there some other way I should be reading the state?

Don't know.

Leave blank for maintainers to fill out if you are uncertain about this.

aestene:
Also sets retain to true for the state topic.
Closes #324 and closes #323


tsundvoll (approved):
LGTM

GodVenn (approved):
LGTM

PR #337, opened 2022-10-12
Log warning on missing inspection data

### LINKED ISSUE #336: Empty inspection list is not handled in storage handler
**Describe the bug**
When the monitor step attempts to retrieve inspections for a given step in a mission to put them in the upload queue, and it gets no inspections, there are no warnings being triggered.
The relevant code is found in [monitor.py](https://github.com/equinor/isar/tree/main/src/isar/state_machine/states/monitor.py) line 79.

**To Reproduce**
Run a mission in which the inspection data is faulty.

**Expected behavior**
We expect a logged warning when there is no inspection data where we expected there to be some.

**Screenshots**

tsundvoll:
Fixes #336.

### DIFF
--- a/src/isar/state_machine/states/monitor.py
+++ b/src/isar/state_machine/states/monitor.py
@@ -85,6 +85,11 @@ class Monitor(State):
             )
             return
 
+        if not inspections:
+            self.logger.warning(
+                f"No inspection data retrieved for step {str(current_step.id)[:8]}"
+            )
+
         # A deepcopy is made to freeze the metadata before passing it to another thread
         # through the queue
         mission_metadata: MissionMetadata = deepcopy(


oysand (approved):
LGTM

src/isar/state_machine/states/monitor.py
```diff
@@ -85,6 +85,11 @@ def _queue_inspections_for_upload(self, current_step: InspectionStep):
             )
             return
 
+        if len(inspections) == 0:
```
aestene, line 88:
```suggestion
        if not inspections:
```

aestene: More the "python" way of doing it I think.

aestene (approved):
One comment you may fix otherwise good.

src/isar/state_machine/states/monitor.py
tsundvoll: Good point, fixed.

PR #344, opened 2022-10-17
Lock uvicorn to 0.0.0.0

oysand:
since the API_HOST environment variable is used to report where one can find isar and uvicorn needed it to be 0.0.0.0


tsundvoll:
Do we also need to change the default value for API_HOST in the settings.py?

tsundvoll (approved):
LGTM

PR #350, opened 2022-10-18
Add role based access control to all endpoints

### LINKED ISSUE #348: Add role based acccess control for ISAR
ISAR currently accepts all requests as long as the user is **authenticated** and is in the Equinor Azure AD group.
We would like to regulate who in our organization can control missions directly to ISAR and for this we need Role Based Access Control, where we can create roles in the `App Registration` and assign them in the `Enterprise Application`.

Role based access control with one role (for now) called "Mission Control"

It will improve security by limiting access to those we grant it to.

GodVenn:
Closes #348 
To test: Pull the PR and run it locally.
Authenticate and test an endpoint.
You should get unauthorized because I am currently the only one with the correct role.
Before approval I will give the role to the team and you should be able to run it, then we can merge.

### DIFF
--- a/src/isar/apis/security/authentication.py
+++ b/src/isar/apis/security/authentication.py
@@ -1,5 +1,9 @@
 import logging
 
+
+from fastapi import Depends
+from fastapi_azure_auth.exceptions import InvalidAuth
+from fastapi_azure_auth.user import User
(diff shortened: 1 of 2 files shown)


Christdej (approved):
LGTM

PR #353, opened 2022-10-19
Add responses to pause, resume, stop mission endpoints

### LINKED ISSUE #351: Empty response body on OK request in scheduling controller
**Describe the bug**
Pause-, resume- and stop mission endpoints in scheduling controller returns "null" in response body for response 200. Mission is successfully pause, stopped or resumed, but the backend throws an error because of the empty response body.

**To Reproduce**
Schedule a mission in frontend -> pause, resume or stop mission and see error messages.

**Expected behavior**
Scheduling api should not return an empty response body on successful stop, pause or resume.

**Screenshots**
image

eivindsjovold:
Closes #351.

### DIFF
--- a/src/isar/apis/models/models.py
+++ b/src/isar/apis/models/models.py
@@ -24,10 +24,15 @@ class StartMissionResponse(BaseModel):
 class PauseMissionResponse(BaseModel):
     mission_id: Union[UUID, int, str, None]
     mission_status: str
-    task_id: str
+    task_id: Union[UUID, int, str, None]
     task_status: str
-    step_id: str
+    step_id: Union[UUID, int, str, None]
     step_status: str
+
+
+class ResumeMissionResponse(BaseModel):
+    mission_id: Union[UUID, int, str, None]
+    mission_status: str
 
 
 # We need to specify our own position/orientation/pose classes that do not contain
--- a/src/isar/apis/models/models.py
+++ b/src/isar/apis/models/models.py
@@ -21,18 +21,13 @@ class StartMissionResponse(BaseModel):
     tasks: List[TaskResponse]
 
 
-class PauseMissionResponse(BaseModel):
+class ControlMissionResponse(BaseModel):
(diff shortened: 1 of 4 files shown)


src/isar/apis/schedule/scheduling_controller.py
```diff
@@ -313,13 +320,14 @@ def stop_mission(self, response: Response):
             return error_message
 
         try:
-            self.scheduling_utilities.stop_mission()
+            stopped_mission_response = self.scheduling_utilities.stop_mission()
```
aestene, line 323: Typehinting

src/isar/apis/models/models.py
```diff
@@ -21,6 +21,15 @@ class StartMissionResponse(BaseModel):
     tasks: List[TaskResponse]
 
 
+class ControlMissionResponse(BaseModel):
+    mission_id: Union[UUID, int, str, None]
+    mission_status: str
+    task_id: Union[UUID, int, str, None]
+    task_status: str
+    step_id: Union[UUID, int, str, None]
+    step_status: str
```
GodVenn, line 30: If the mission/task/sted Id's can be "None", should the statuses not also have the option to be "None"?

GodVenn (commented):
Looks good, please see comment

phhuseboe (approved):
LGTM

src/isar/apis/models/models.py
aestene: This is a problem with the mission id setup and should be solved there really. The mission ID should not be a union but simply a UUID. I suggest we solve that instead. Mainly it could be solved by setting the mission IDs in the predefined missions to UUIDs and then I think we should be golden. Then every place with a Union like this could be just UUID.

aestene: Also they can't really be `None` in this place. It is only a workaround for something occurring when setting up the mission itself so it is not required to have `None` for the statuses.

PR #358, opened 2022-11-03
Update MQTT communication in ReadMe

EinarUeland:


README.md
```diff
@@ -342,14 +342,17 @@ turned off but may be activated by setting the environment variable
 ISAR_MQTT_ENABLED = true
 ```
 
-The connection to the broker will be determined by the following configuration values which may all be overwritten
-through the environment.
+The connection to the broker will be determined by the following configuration values specified in ```settings.env```
```
GodVenn, line 345: I think the previous text is better here, as the variables have default values defined in the [settings.py](https://github.com/equinor/isar/tree/main/src/isar/config/settings.py) file

README.md
```diff
@@ -342,14 +342,17 @@ turned off but may be activated by setting the environment variable
...
 
 ```
 ISAR_MQTT_USERNAME
 ISAR_MQTT_HOST
 ISAR_MQTT_PORT
 ```
 
-In addition, the `ISAR_MQTT_PASSWORD` environment variable should be available for connection to the broker. If username
-and password is not specified both will default to empty strings.
+To specify broker password, setup a .env file with the following environment variable:
```
GodVenn, line 353:
```suggestion
To specify broker password, add the following environment variable to a .env file in the root of the repository:
```

GodVenn (commented):
Good initiative

README.md
EinarUeland: In practice default values are specified in settings.env since the environment overwrite settings.py, but I will modify to include both in text.

GodVenn (approved):
LGTM

PR #365, opened 2022-11-15
Configure pipeline to not cancel jobs on fail

EinarUeland:
Previously the pipeline cancelled all python version jobs if one version failed. With this modification, one can see exactly which version fails.


tsundvoll (approved):
LGTM :+1:

PR #367, opened 2022-11-23
Retry on failed attempt not on queue loop

aestene:
The `handle_stop_fail` function should not be called on every thread not finished error as this is basically the while loop but rather if we fail to stop.


tsundvoll (approved):
LGTM :+1:

PR #369, opened 2022-11-25
Correct logging when mission is accepted

anetteu:
A mission that is recieved, accepted, and initiated by ISAR is not necessarily started by the robot.


GodVenn:
But isn't it started in ISAR?

GodVenn (dismissed):
I agree that it was confusing before, but "start" and "initiate" have the same meanings.
And I think the error message should say failed to start instead of accept, since with accept it may sound like the mission is in the wrong format or contains invalid information, which is not the case with QueueTimeouts.

aestene:

aestene (approved):
LGTM

PR #371, opened 2022-12-01
Fix dockerfile

tsundvoll:


src/isar/mission_planner/local_planner.py
```diff
@@ -29,7 +29,7 @@ def get_mission(self, mission_id) -> Mission:
             mission: Mission = missions[mission_id]["mission"]
             mission.set_unique_id_and_metadata()
             return mission
-        except Exception as e:
+        except KeyError as e:
```
Christdej, line 32: Do we still need to catch general exception and throw error accordingly?

Christdej (changes requested):

Christdej (approved):
LGTM

src/isar/mission_planner/local_planner.py
tsundvoll: Added catching the general exception in addition

PR #377, opened 2023-01-04
Add enpoint for basic info

monkeyysups:


src/isar/apis/api.py
```diff
@@ -222,6 +222,14 @@ def _create_scheduler_router(self) -> APIRouter:
             },
         )
 
+        router.add_api_route(
```
Christdej, line 225: At the top of this function we create the router:
`router: APIRouter = APIRouter(tags=["Scheduler"])`

I think that this information endpoint should be in a seperate router as this has nothing to do with scheduling.
For instance create a new function `def _create_info_router(self) -> APIRouter:`
Then create a new router and add this endpoint to that router

Christdej (changes requested):

src/isar/apis/api.py
monkeyysups: fixed

Christdej (approved):
LGTM

PR #378, opened 2023-01-04
State machine diagram and small state machine fixes

### LINKED ISSUE #328: Why does stop_mission require not being in the initializing state?

To use the HTTP API correctly, it seems that you need to know the state (`Idle`, `Monitor`, etc.). Some API calls are only allowed/available when in some specifc states. From how I read the code (https://github.com/equinor/isar/blob/4699f55f6f012da44bf30bd56398d923d943984b/src/isar/apis/schedule/scheduling_controller.py) understand the constraints are:

 * `start_mission_by_id`: must be in state `Off` or `Idle`.
 * `start_mission`: must be in state `Off` or `Idle`.
 * `pause_mission`: must be in state `InitiateStep`, `Monitor` or `Off`.
 * `resume_mission`: must be in state `Off` or `Paused`.
 * `stop_mission`: must be in state `Off`, `InitiateStep`, `Monitor`, `Paused`, `StopStep`.
 * `drive_to`: must be in state `Off` or `Idle`.

Unless this is already available somehow, I suggest to add something like this list to the static docs or the interactive API docs. 

Some questions:

 * Will the API calls work in the `Off` state? If not, is there a correct HTTP response for this?
 * Why not `pause_mission` in `StopStep` or `Initialize`?
 * Why not `stop_mission` in `Initialize`?

Christdej:
Closes #328


tsundvoll (changes requested):
For the state checks, is it possible to check the condition of which state the FSM is not in, i.e. the negated condition? Could this make it more readable?

MortFred (approved):
LGTM

Christdej:
> For the state checks, is it possible to check the condition of which state the FSM is not in, i.e. the negated condition? Could this make it more readable?

Yes, I think this might make it more readable, will change and we can see if we like it better

src/isar/apis/api.py
```diff
@@ -93,6 +93,7 @@ def _create_scheduler_router(self) -> APIRouter:
             "/schedule/start-mission/{id}",
             self.scheduling_controller.start_mission_by_id,
             methods=["POST"],
+            deprecated=True,
```
Christdej, line 96: @tsundvoll @aestene What do you guys think about this? End point still usable, but If I understand correct we will not use this for other then testing locally?

aestene: I think it works well until we decide to remove it.

tsundvoll (approved):
LGTM

PR #379, opened 2023-01-04
Added thread for inspection

monkeyysups:


Christdej (changes requested):
Not the original scope of the issue, but can you also add names to the threads we create, for easier debugging. In `class ThreadedRequest:` we create a `Thread(target=self._thread_func, args=request_args)` these Threads have the possibility to be named.

src/isar/services/utilities/threaded_request.py
```diff
@@ -11,11 +11,12 @@ def __init__(self, request_func: Any) -> None:
         self._exception: Optional[Exception] = None
         self._exception_lock: Lock = Lock()
 
-    def start_thread(self, *request_args) -> bool:
+    def start_thread(self, *request_args, thread_name: str) -> bool:
         if self._is_thread_alive():
             return False
         self._output = None
         self._thread = Thread(target=self._thread_func, args=request_args)
+        self._thread.name = thread_name
```
Christdej, line 19: We could make the thread name optional with:
```suggestion
    def start_thread(self, *request_args, **kwargs) -> bool:
        if self._is_thread_alive():
            return False
        self._output = None
        self._thread = Thread(target=self._thread_func, args=request_args, **kwargs)
        self._thread.name = thread_name
```
Then in places we call this function you can simply do:
```
get_inspections_thread = ThreadedRequest(self._process_finished_step)
get_inspections_thread.start_thread(
    self.state_machine.current_step, name="get_inspections"
)
```
And this will get passed over to the creation of the Thread

Christdej (changes requested):

src/isar/state_machine/states/initiate_step.py
```diff
@@ -65,7 +65,10 @@ def _run(self):
                 self.initiate_step_thread = ThreadedRequest(
                     self.state_machine.robot.initiate_step
                 )
-                self.initiate_step_thread.start_thread(self.state_machine.current_step)
+                self.initiate_step_thread.start_thread(
+                    self.state_machine.current_step,
+                    name="state_machine_curent_step",
```
aestene, line 70:
```suggestion
                    name="State Machine Initiate Step",
```

src/isar/state_machine/states/monitor.py
```diff
@@ -52,7 +52,7 @@ def _run(self):
                 self.step_status_thread = ThreadedRequest(
                     self.state_machine.robot.step_status
                 )
-                self.step_status_thread.start_thread()
+                self.step_status_thread.start_thread(name="step_status")
```
aestene, line 55:
```suggestion
                self.step_status_thread.start_thread(name="State Machine Monitor Current Step")
```

src/isar/state_machine/states/monitor.py
```diff
@@ -65,7 +65,10 @@ def _run(self):
             self.state_machine.current_step.status = step_status
 
             if self._step_finished(step=self.state_machine.current_step):
-                self._process_finished_step(step=self.state_machine.current_step)
+                get_inspections_thread = ThreadedRequest(self._process_finished_step)
+                get_inspections_thread.start_thread(
+                    self.state_machine.current_step, name="get_inspections"
+                )
```
aestene, line 71:
```suggestion
                get_inspections_thread.start_thread(
                    self.state_machine.current_step, name="State Machine Get Inspections"
                )
```

src/isar/state_machine/states/stop_step.py
```diff
@@ -37,7 +37,7 @@ def _run(self):
         while True:
             if not self.stop_thread:
                 self.stop_thread = ThreadedRequest(self.state_machine.robot.stop)
-                self.stop_thread.start_thread()
+                self.stop_thread.start_thread(name="stopper")
```
aestene, line 40:
```suggestion
                self.stop_thread.start_thread(name="State Machine Stop Robot")
```

aestene (changes requested):
Suggested changes to names to better reflect functionality and naming conventions for other threads in the system.

Christdej (approved):
LGTM

aestene (approved):

Issue #387, opened 2023-01-13
Fastapi 0.89.1 fails


tsundvoll:
**Describe the bug**
```
venv-robot/lib/python3.10/site-packages/fastapi/routing.py:255: in app
    content = await serialize_response(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    async def serialize_response(
        *,
        field: Optional[ModelField] = None,
        response_content: Any,
        include: Optional[Union[SetIntStr, DictIntStrAny]] = None,
        exclude: Optional[Union[SetIntStr, DictIntStrAny]] = None,
        by_alias: bool = True,
        exclude_unset: bool = False,
        exclude_defaults: bool = False,
        exclude_none: bool = False,
        is_coroutine: bool = True,
    ) -> Any:
        if field:
            errors = []
            response_content = _prepare_response_content(
                response_content,
                exclude_unset=exclude_unset,
                exclude_defaults=exclude_defaults,
                exclude_none=exclude_none,
            )
            if is_coroutine:
                value, errors_ = field.validate(response_content, {}, loc=("response",))
            else:
                value, errors_ = await run_in_threadpool(
                    field.validate, response_content, {}, loc=("response",)
                )
            if isinstance(errors_, ErrorWrapper):
                errors.append(errors_)
            elif isinstance(errors_, list):
                errors.extend(errors_)
            if errors:
>               raise ValidationError(errors, field.type_)
E               pydantic.error_wrappers.ValidationError: <exception str() failed>
```

**To Reproduce**
Run `pytest` with fastapi==0.89.1

**Expected behavior**
All tests pass. This works with fastpi==0.88.0

**Screenshots**
N/A

tsundvoll:
- [ ] Let each endpoint return just one type, and raise the exception (instead of returning error message)

Possible refactorings:
- [ ] Have similar responses in the Router setup, instead of in each route

oysand:
Is this fixed in #388 ?

tsundvoll:
Yes, solved by #388

PR #388, opened 2023-01-13
Bugfix fastapi type in test

tsundvoll:


tsundvoll:
Resolves #387

src/isar/services/utilities/scheduling_utilities.py
```diff
@@ -24,28 +30,192 @@ class SchedulingUtilities:
...
+                f"Bad Request - Robot is not capable of performing mission."
+                f" Missing functionalities: {missing_capabilities}."
+            )
+            self.logger.error(error_message)
+            raise HTTPException(
+                status_code=HTTPStatus.BAD_REQUEST,
+                detail=error_message,
+            )
+
+        assert is_capable
```
oysand, line 120: Should we catch and use AssertionError?

src/isar/services/utilities/scheduling_utilities.py
```diff
@@ -24,28 +30,192 @@ class SchedulingUtilities:
...
+
+    def assert_state_machine_is_idle(self, state: States):
```
oysand, line 122: Check if state machine is ready to receive a new mission rather than checking if it is idle

tests/isar/apis/scheduler/test_scheduler_router.py
```diff
@@ -0,0 +1,291 @@
...
+    @mock.patch.object(SchedulingUtilities, "get_state", mock_return_idle)
+    def test_state_machine_in_conflicting_state(self, client: TestClient):
+        response = client.post(url=self.schedule_resume_mission_path)
+        assert response.status_code == HTTPStatus.CONFLICT
+        assert response.json() == {
+            "detail": "Conflict - Resume command received in invalid state - State: idle"
+        }
+
+    @mock.patch.object(SchedulingUtilities, "get_state", mock_return_paused)
+    @mock.patch.object(SchedulingUtilities, "_send_command", mock_QueueTimeoutError)
```
oysand, line 212: Mixing snake and camel case

tests/isar/services/utilities/test_scheduling_utilities.py
```diff
@@ -1,16 +1,52 @@
...
-import queue
-from queue import Queue
+@pytest.mark.parametrize(
+    "capabilities, expectation",
+    [
+        (["drive_to_pose"], pytest.raises(HTTPException)),
+        (["drive_to_pose", "take_image"], does_not_raise()),
+    ],
+)
+def test_assert_robot_is_capable_of_mission(
```
oysand, line 31: Consider splitting into two test to avoid complexity from does_not_raise()

tests/isar/services/utilities/test_scheduling_utilities.py
```diff
@@ -1,16 +1,52 @@
...
+
+
+@pytest.mark.parametrize(
+    "state, expectation",
+    [
+        (States.Initialize, pytest.raises(HTTPException)),
+        (States.Idle, does_not_raise()),
+    ],
+)
+def test_assert_state_machine_is_idle(
```
oysand, line 48: Consider splitting into two test to avoid complexity from does_not_raise()

src/isar/services/utilities/scheduling_utilities.py
tsundvoll: Apparently, this is a bad idea, since assertions can be optimized away:
https://realpython.com/python-assert-statement/#when-not-to-use-assertions

tsundvoll: Renaming the functions to `verify_...` and returning the boolean instead of asserting

tests/isar/apis/scheduler/test_scheduler_router.py
tsundvoll: Good catch

tests/isar/services/utilities/test_scheduling_utilities.py
tsundvoll: Good idea, implemented

tests/isar/services/utilities/test_scheduling_utilities.py
tsundvoll: Good idea, implemented

src/isar/services/utilities/scheduling_utilities.py
tsundvoll: Good idea, implemented

src/isar/services/utilities/scheduling_utilities.py
```diff
@@ -24,28 +30,192 @@ class SchedulingUtilities:
...
+                StartMissionMessage(
+                    mission=deepcopy(mission), initial_pose=initial_pose
+                ),
+                self.queues.start_mission,
+            )
+        except QueueTimeoutError:
+            error_message = "Timeout - Failed to start mission in ISAR"
+            self.logger.error(error_message)
+            raise HTTPException(
+                status_code=HTTPStatus.REQUEST_TIMEOUT, detail=error_message
```
GodVenn, line 157: 408 Request timeout implies a user error (all 400 codes do), I believe this should be a 500 internal server error as it has nothing to do with the user input

src/isar/services/utilities/scheduling_utilities.py
```diff
@@ -24,28 +30,192 @@ class SchedulingUtilities:
...
+        ------
+        HTTTPException 408 Request timeout
+            If there is a timeout while communicating with the state machine
+        """
+        try:
+            return self._send_command(True, self.queues.pause_mission)
+        except QueueTimeoutError:
+            error_message = "Timeout - Failed to pause mission"
+            self.logger.error(error_message)
+            raise HTTPException(
```
GodVenn, line 174: 408 Request timeout implies a user error (all 400 codes do), I believe this should be a 500 internal server error as it has nothing to do with the user input

src/isar/services/utilities/scheduling_utilities.py
```diff
@@ -24,28 +30,192 @@ class SchedulingUtilities:
...
+        ------
+        HTTTPException 408 Request timeout
+            If there is a timeout while communicating with the state machine
+        """
+        try:
+            return self._send_command(True, self.queues.resume_mission)
+        except QueueTimeoutError:
+            error_message = "Timeout - Failed to resume mission"
+            self.logger.error(error_message)
+            raise HTTPException(
```
GodVenn, line 193: 408 Request timeout implies a user error (all 400 codes do), I believe this should be a 500 internal server error as it has nothing to do with the user input

src/isar/services/utilities/scheduling_utilities.py
```diff
@@ -24,28 +30,192 @@ class SchedulingUtilities:
...
+            If there is a timeout while communicating with the state machine
+        """
+        try:
+            stop_mission_response: ControlMissionResponse = self._send_command(
+                True, self.queues.stop_mission
+            )
+        except QueueTimeoutError:
+            error_message = "Timeout - Failed to stop mission"
+            self.logger.error(error_message)
+            raise HTTPException(
```
GodVenn, line 214: 408 Request timeout implies a user error (all 400 codes do), I believe this should be a 500 internal server error as it has nothing to do with the user input

GodVenn (changes requested):
Looks good, I just think those status codes should be changed

src/isar/services/utilities/scheduling_utilities.py
tsundvoll: Good catch, changed to 500 Internal server error

src/isar/services/utilities/scheduling_utilities.py
tsundvoll: Good catch, changed to 500 Internal server error

src/isar/services/utilities/scheduling_utilities.py
tsundvoll: Good catch, changed to 500 Internal server error

src/isar/services/utilities/scheduling_utilities.py
tsundvoll: Good catch, changed to 500 Internal server error

GodVenn (approved):
LGTM

PR #390, opened 2023-01-13
Publish robot availability as keep alive signal

### LINKED ISSUE #341: Publish isar status over mqtt
ISAR should periodically update its status over mqtt.

If you have an idea of how to implement the feature please describe it here.

N/A

aeshub:
Closes #341

### DIFF
--- a/src/isar/config/settings.py
+++ b/src/isar/config/settings.py
@@ -307,7 +307,7 @@ class RobotSettings(BaseSettings):
 
     # Model of the robot which ISAR is connected to
     # This should be set in the robot package settings.env file
-    ROBOT_MODEL: RobotModel = Field(default=RobotModel.Robot)
+    ROBOT_MODEL: RobotModel = Field(default=RobotModel.Robot)  # type: ignore
 
     class Config:
         env_file_encoding = "utf-8"
--- a/src/robot_interface/telemetry/payloads.py
+++ b/src/robot_interface/telemetry/payloads.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass
 from datetime import datetime
-from typing import List
+from typing import List, Union
 from uuid import UUID
 
 from alitra import Pose
@@ -35,7 +35,7 @@ class RobotStatusPayload:
     robot_name: str
     robot_status: RobotStatus
     current_isar_state: State
-    current_mission_id: str
+    current_mission_id: Union[UUID, int, str, None]
     current_task_id: UUID
     current_step_id: UUID
     timestamp: datetime
--- a/src/isar/config/keyvault/keyvault_service.py
+++ b/src/isar/config/keyvault/keyvault_service.py
@@ -1,5 +1,6 @@
 import logging
 import traceback
+from typing import Union
 
 from azure.core.exceptions import (
     ClientAuthenticationError,
@@ -52,16 +53,17 @@ class Keyvault:
             traceback.print_exc()
             raise KeyvaultError  # type: ignore
 
-    def get_secret_client(self):
+    def get_secret_client(self) -> SecretClient:
         try:
+            credential: Union[ClientSecretCredential, DefaultAzureCredential]
             if self.client_id and self.client_secret and self.tenant_id:
-                credential: ClientSecretCredential = ClientSecretCredential(
+                credential = ClientSecretCredential(
                     tenant_id=self.tenant_id,
                     client_id=self.client_id,
                     client_secret=self.client_secret,
                 )
             else:
-                credential: DefaultAzureCredential = DefaultAzureCredential()
+                credential = DefaultAzureCredential()
         except ClientAuthenticationError:
             self.logger.error("Failed to authenticate to Azure.")
             traceback.print_exc()
--- a/src/isar/config/log.py
+++ b/src/isar/config/log.py
@@ -8,7 +8,7 @@ from uvicorn.logging import ColourizedFormatter
 from isar.config.settings import settings
 
 
-def setup_logger():
+def setup_logger() -> None:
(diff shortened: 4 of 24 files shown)


src/robot_interface/telemetry/payloads.py
```diff
@@ -16,3 +19,25 @@ class TelemetryBatteryPayload:
     battery_level: float
     robot_id: str
     timestamp: datetime
+
+
+@dataclass
+class VideoStream:
+    name: str
+    url: str
```
Christdej, line 27: Might want to move this to the get information endpoint, so we dont have to publish this all the time

src/robot_interface/telemetry/payloads.py
```diff
@@ -16,3 +19,25 @@ class TelemetryBatteryPayload:
...
+    serial_number: str
+    video_streams: List[VideoStream]
+    host: str
+    port: int
+    robot_available: bool
+    isar_available: bool
+    current_state: State
+    current_mission_id: str
+    current_task_id: UUID
+    current_step_id: UUID
```
Christdej, line 43: Some of these properties might be moved to get information endpoint so we dont publish alot of unnecessary info

Christdej (changes requested):

src/robot_interface/robot_interface.py
```diff
@@ -136,3 +136,19 @@ def get_telemetry_publishers(self, queue: Queue, robot_id: str) -> List[Thread]:
 
         """
         raise NotImplementedError
+
+    @abstractmethod
+    def is_robot_online(self) -> bool:
```
oysand, line 141: Would it be better to have an indication of robot status? That way we can get a response indicating Offline or not, but also incorporate Available and Busy. It may be that the robot is online, but unavalable

src/robot_interface/telemetry/payloads.py
aestene: Ended up with a solution where the info is split into two different topics such that it is a bit more natural.

src/robot_interface/robot_interface.py
aestene: Yes, implemented an enum which reflects the status from Flotilla.

main.py
```diff
@@ -51,13 +58,27 @@
         mqtt_thread: Thread = Thread(
             target=mqtt_client.run, name="ISAR MQTT Client", daemon=True
         )
-
         threads.append(mqtt_thread)
 
+        robot_status: RobotStatusPublisher = RobotStatusPublisher(
```
oysand, line 63: Worth adding *_publisher to this and and *Publisher objects below?

src/robot_interface/telemetry/payloads.py
```diff
@@ -16,3 +21,31 @@ class TelemetryBatteryPayload:
...
+
+
+@dataclass
+class VideoStream:
+    name: str
+    url: str
+
+
+@dataclass
+class RobotStatusPayload:
```
oysand, line 33: Should we not include current_isar_state, current_mission_id, current_task_id and current_step_id until we need it for something? That way we can add it as needed and only supply the information that Flotilla actually use

src/isar/config/settings.py
```diff
@@ -179,6 +186,25 @@ class Settings(BaseSettings):
     # Name or unique ID of robot
     ROBOT_ID: str = Field(default="R2-D2")
```
Christdej, line 187: Move this to RobotSettings and change default value to "default-robot"?

main.py
aestene: Done.

src/robot_interface/models/mission/status.py
```diff
@@ -7,3 +7,9 @@ class StepStatus(str, Enum):
     InProgress: str = "in_progress"
     Failed: str = "failed"
     Cancelled: str = "cancelled"
+
+
+class RobotStatus(Enum):
+    Available: str = "available"
+    Busy: str = "busy"
+    Offline: str = "offline"
```
Christdej, line 15: Do we want to add "Deprecated" here aswell as it is comming in Flotilla soon?

Christdej (changes requested):

src/robot_interface/telemetry/payloads.py
aestene: I would argue this message is "general" in the sense that it's the status of the ISAR instance. It is not a message that should exist for Flotilla because they are modular applications. It's Flotilla's choice to use this information for what it needs it to.

If other applications should use ISAR this information is the status. Also, it might be good to have it for the timeseries logging. Unless we go with seperate publishers for each one.

src/isar/config/settings.py
aestene: I started working on it but ended up deciding it is the next pull request. Mainly because it required a lot of change in Flotilla and I don't want this PR to grow any further.

Christdej: Yeah, I agreed

src/robot_interface/models/mission/status.py
aestene: Maybe, it depends what deprecated means for ISAR. Is the ISAR instance still running? Probably not. 

In any case I think we can wait with adding it until required in this case.

Christdej: Yeah, no I dont think ISAR ever needs to be "Deprecated".

Christdej (approved):
LGTM

src/robot_interface/telemetry/payloads.py
oysand: Agreed

oysand (approved):
LGTM

PR #400, opened 2023-02-08
Update SLIMM endpoint and add analysis type

### LINKED ISSUE #368: Remove 'start_time' in 'InspectionMetadata'
Remove the 'start_time' attribute from the 'InspectionMetadata' class in 'src/robot_interface/models/inspection/inspection.py'. Time is anyways baked into the 'time_indexed_pose' attribute. Remember to update everywhere the class is used, also in the different robot repositories.

I can't see that start time is used anywhere now, so it will not break any existing functionality.

### LINKED ISSUE #385: Add optional field to InspectionStep to include analysis type
This should be further used to determine which analysis pipelines to run later. Thus it should be included as metadata to any data that is uploaded.

The `InspectionStep` object should include a field which is optional that specifies what analysis should run on that inspection. This might be a list of different analysis options. 

The local planner should support adding analysis types as part of the JSON file which should then be propagated to the InspectionStep and then to the metadata upon upload. 

N/A

### LINKED ISSUE #386: Add optional field for tag in predefined missions used by local planner
This should function the same as for the Echo planner where the tag is propagated through the `Task` object. 

Add option of passing metadata with the local planner. This issue may be seen together with #385 where the analysis type should be added to the planner too.

N/A

Christdej:
Closes #385 
Closes #386

Needs to be done together with:
https://github.com/equinor/isar-robot/pull/56

### DIFF
--- a/src/isar/apis/models/start_mission_definition.py
+++ b/src/isar/apis/models/start_mission_definition.py
@@ -1,4 +1,5 @@
-from typing import List, Optional, Type
+from enum import Enum
+from typing import List, Optional
 
 from alitra import Position
 from pydantic import BaseModel, Field
@@ -9,29 +10,32 @@ from isar.models.mission.mission import Mission, Task
 from robot_interface.models.mission.step import (
     STEPS,
     DriveToPose,
-    InspectionStep,
     TakeImage,
     TakeThermalImage,
     TakeThermalVideo,
     TakeVideo,
 )
 
-inspection_step_types: List[Type[InspectionStep]] = [
-    TakeImage,
-    TakeThermalImage,
-    TakeVideo,
-    TakeThermalVideo,
-]
+
+class InspectionTypes(str, Enum):
+    image = "Image"
+    thermal_image = "ThermalImage"
+    video = "Video"
+    thermal_video = "ThermalVideo"
+
+
+class StartMissionInspectionDefinition(BaseModel):
+    type: InspectionTypes = Field(default=InspectionTypes.image)
+    inspection_target: InputPosition
+    analysis_types: Optional[List]
+    duration: Optional[float]
+    metadata: Optional[dict]
 
 
 class StartMissionTaskDefinition(BaseModel):
     pose: InputPose
     tag: Optional[str]
-    inspection_target: InputPosition
-    inspection_types: List[str] = Field(
-        default=[step.get_inspection_type().__name__ for step in inspection_step_types]
-    )
-    video_duration: Optional[float] = Field(default=10)
+    inspections: List[StartMissionInspectionDefinition]
 
 
 class StartMissionDefinition(BaseModel):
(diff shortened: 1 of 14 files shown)


Christdej:
Will require updates in isar-robots and flotilla

src/isar/apis/models/start_mission_definition.py
```diff
@@ -9,29 +10,32 @@
...
 
 class StartMissionTaskDefinition(BaseModel):
     pose: InputPose
     tag: Optional[str]
-    inspection_target: InputPosition
-    inspection_types: List[str] = Field(
-        default=[step.get_inspection_type().__name__ for step in inspection_step_types]
-    )
-    video_duration: Optional[float] = Field(default=10)
+    inspections: List[StartMissionInspectionDefinition]
```
GodVenn, line 38: Won't the inspection target always be the same for each inspection in a task? Because there is one tag per task?
Should it therefore be outside of the `StartMissionInspectionDefinition` ?

src/isar/config/settings.py
```diff
@@ -151,14 +151,14 @@ class Settings(BaseSettings):
...
-    SLIMM_CLIENT_ID: str = Field(default="94c048cc-58e9-4570-85c0-4028c50ab6f3")
+    SLIMM_CLIENT_ID: str = Field(default="c630ca4d-d8d6-45ab-8cc6-68a363d0de9e")
 
     # Scope for access to SLIMM Ingestion API
     SLIMM_APP_SCOPE: str = Field(default=".default")
 
     # URL for SLIMM endpoint
     SLIMM_API_URL: str = Field(
-        default="https://slimmingestapitest.azurewebsites.net/SpatialIngest"
+        default="https://scinspectioningestapitest.azurewebsites.net/Ingest"
```
GodVenn, line 161: This URL has "apiTEST" in it, do they have a production API for us to use?

src/isar/storage/slimm_storage.py
```diff
@@ -30,41 +31,61 @@ def __init__(self, request_handler: RequestHandler) -> None:
...
-        inspection_type: str = get_inspection_type(inspection=inspection)
+    def store(self, inspection: Inspection, metadata: MissionMetadata):
         filename: str = get_filename(
             mission_id=metadata.mission_id,
-            inspection_type=inspection_type,
+            inspection_type=type(inspection).__name__,
             inspection_id=inspection.id,
         )
-        filename_with_ending: str = f"{filename}.{inspection.metadata.file_type}"
+        filename = f"{filename}.{inspection.metadata.file_type}"
```
GodVenn, line 40: typehinting missing (: str)

src/isar/storage/utilities.py
```diff
@@ -14,12 +14,10 @@
 def construct_local_paths(
     inspection: Inspection, metadata: MissionMetadata
 ) -> Tuple[Path, Path]:
-    inspection_type: str = get_inspection_type(inspection=inspection)
-
     folder: Path = Path(str(metadata.mission_id))
     filename: str = get_filename(
         mission_id=metadata.mission_id,
-        inspection_type=inspection_type,
+        inspection_type=inspection.get_metadata_type_str(),
```
GodVenn, line 20: Should this be metadata type or inspection type?

GodVenn (commented):
This is a nice improvement

src/isar/apis/models/start_mission_definition.py
Christdej: Had a discussion with others on the team, but concluded that having the possibility to adjust target but standing on the same place might be beneficial, especially when inspecting a pipe or larger tags

src/isar/config/settings.py
Christdej: Not yet, because this is a test feature that is not yet in production

src/isar/storage/slimm_storage.py
Christdej: Nope, because `filename` is defined on line 35, we are just reusing the variable and hence do not need to typehint it again

src/isar/storage/utilities.py
Christdej: yes! Good catch

GodVenn (approved):
LGTM

PR #402, opened 2023-02-10
Log to application insights

### LINKED ISSUE #382: Send log messages to Application Insight
* All log messages that are logged to console should also be sent to Application Insight and stored.

Include possibility to log based on log level.

GodVenn:
Closes #382 
We might need to consider linking ISAR to its keyvault for better management of the connection string.
Currently there needs to exist an environment variable called `APPLICATIONINSIGHTS_CONNECTION_STRING` containing the connection string for the correct Application Insights in Azure.

Currently logging to:
https://portal.azure.com/#@StatoilSRM.onmicrosoft.com/resource/subscriptions/c389567b-2dd0-41fa-a5da-d86b81f80bda/resourceGroups/IsarDev/providers/Microsoft.Insights/components/IsarDevAI/overview

Go to Logs and query `traces` to see log messages.
Otherwise, you can find the request (API -request/responses) logging in the overview.


src/isar/apis/api.py
```diff
@@ -5,17 +5,27 @@
...
 import click
 import uvicorn
-from fastapi import FastAPI, Security
+from fastapi import FastAPI, Security, Request
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.routing import APIRouter
 from injector import inject
 from pydantic import AnyHttpUrl
 
+# Opencensus imports
```
Christdej, line 14: Unnecessary comment in my opinion

src/isar/config/log.py
```diff
@@ -1,32 +1,45 @@
...
 
-    for loggers in log_config["loggers"].keys():
-        logging.getLogger(loggers).addHandler(log_handler)
-        logging.getLogger(loggers).setLevel(log_levels[loggers])
-    logging.getLogger().addHandler(log_handler)
+def configure_azure_handler(log_config: dict) -> logging.Handler:
+    # Automatically gets connection string from env variable 'APPLICATIONINSIGHTS_CONNECTION_STRING'
+    handler = AzureLogHandler()
+    handler.setLevel(log_config["root"]["level"])
+    return handler
```
Christdej, line 45: Add `APPLICATIONINSIGHTS_CONNECTION_STRING` to settings.py with description and a default?

Christdej (changes requested):

src/isar/apis/api.py
```diff
@@ -80,6 +90,26 @@ def _create_app(self) -> FastAPI:
 
         app.include_router(router=self._create_info_router())
 
+        @app.middleware("http")
+        async def middlewareOpencensus(request: Request, call_next):
+            tracer = Tracer(
+                exporter=AzureExporter(),
```
Christdej, line 96:
```suggestion
                exporter=AzureExporter(connection_string=settings.APPLICATIONINSIGHTS_CONNECTION_STRING),
```

src/isar/apis/api.py
GodVenn: Agree

src/isar/config/log.py
GodVenn: Good idea

src/isar/apis/api.py
GodVenn: This one also gets it automatically (like the log handler), should i set both explicitly or neither?

src/isar/config/log.py
GodVenn: Although, it is a secret, should it still be added?

src/isar/apis/api.py
GodVenn: Will set both explicitly

Christdej (approved):
LGTM

PR #407, opened 2023-02-20
Check battery level before initiating step

UsamaEquinorAFK:
Exception added in robot_exceptions.py and
reflected in initiate_step.py.

### DIFF
--- a/src/isar/config/settings.py
+++ b/src/isar/config/settings.py
@@ -1,7 +1,7 @@
 import importlib.resources as pkg_resources
 import os
 from typing import List
-
+from dotenv import load_dotenv
 from pydantic import BaseSettings, Field, validator
 
 from isar.config import predefined_missions
@@ -299,6 +299,7 @@ class Settings(BaseSettings):
         case_sensitive = True
 
 
+load_dotenv()
 settings = Settings()
 
 
--- a/src/isar/state_machine/states/initiate_step.py
+++ b/src/isar/state_machine/states/initiate_step.py
@@ -12,7 +12,9 @@ from isar.services.utilities.threaded_request import (
 from robot_interface.models.exceptions import (
     RobotException,
     RobotInfeasibleStepException,
+    RobotLowBatteryException,
 )
+
 
 if TYPE_CHECKING:
     from isar.state_machine.state_machine import StateMachine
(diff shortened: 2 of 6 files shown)


src/robot_interface/robot_interface.py
```diff
@@ -153,3 +153,13 @@ def robot_status(self) -> RobotStatus:
...
+
+    # @abstractmethod
+    # def is_in_condition_to_perform_mission(self, mission) -> bool:
+
+    #    """
+    #    Called before scheduling a mission, returns true or false, can be called during a mission.
+    #    Import from state machine.py , variable of class mission current mission line 194.
+    #    """
+
+    #    raise NotImplementedError
```
tsundvoll, line 165: This should be removed

src/isar/config/settings.py
```diff
@@ -299,6 +299,7 @@ class Config:
         case_sensitive = True
 
 
+load_dotenv()
```
tsundvoll, line 302: This might makes sense, but what is the reason it is moved from main to here?

src/isar/config/settings.env
```diff
@@ -1,4 +1,4 @@
-ISAR_ROBOT_PACKAGE = isar_robot
+ISAR_ROBOT_PACKAGE = isar_taurob
```
GodVenn, line 1:
```suggestion
ISAR_ROBOT_PACKAGE = isar_robot
```

GodVenn: Default should be isar_robot

GodVenn (commented):
Note on the commits, these 3 commits are essentially one, I think they should be squashed as they solve one thing, not 3.
Also, git commit messages should be in imperative verb tense, e.g. "Check battery" instead of "Checks battery" :) 

This works, but i'm worried about the scalability for other status checks we need to pass before starting missions.
How will pressure be handled? 
And future robot-dependent checks now all have to be defined as exceptions in ISAR, instead of a scalable function that can be defined in each isar-robot repo.

src/isar/config/settings.env
UsamaEquinorAFK: Pressure is not included in this PR. 

"And future robot-dependent checks now all have to be defined as exceptions in ISAR, instead of a scalable function that can be defined in each isar-robot repo." - I reckon that's how it is handled in this PR. 

"These 3 commits are essentially one, I think they should be squashed as they solve one thing, not 3." - I Agree.

src/isar/state_machine/states/initiate_step.py
```diff
@@ -82,6 +84,19 @@ def _run(self) -> None:
...
                 transition = self.state_machine.step_infeasible  # type: ignore
                 break
+
+            except RobotLowBatteryException as e:
+                self.logger.warning(
+                    f"Battery too low to perform step"
+                    f"{type(self.state_machine.current_step).__name__}"
+                    f"Current Battery Level: {str(e.battery_level)}"
+                )
+                print("Sending to Idle: ")
```
aestene, line 94: Use the logger instead of printing here. So `self.logger.info()`

src/isar/config/settings.env
```diff
@@ -24,3 +24,4 @@ ISAR_MQTT_HOST = localhost
 ISAR_MQTT_PORT = 1883
 
 ISAR_KEYVAULT_NAME = IsarDevKv
+
```
aestene, line 27: Remove this

aestene (approved):
LGTM

PR #409, opened 2023-02-23
Mark mission as partially successfull

### LINKED ISSUE #406: Ensure mission is marked as partially successfull if some but not all steps succeeds
**Describe the bug**
Today the mission status is marked as failed immediately when the initial_step_failed is called.

**To Reproduce**
Have one step fail to initialize and another to succeed.

**Expected behavior**
The mission should be marked as partially successful if some but not all steps succeeds.

**Screenshots**
One example that breaks the desired behaviour is this in `src/isar/state_machine/state_machine.py`
```python
    def _initiate_step_failed(self) -> None:
        self.current_step.status = StepStatus.Failed
        self.current_mission.status = MissionStatus.Failed
        self.publish_step_status()
        self._finalize()
```

Also check for other occurrences.

Christdej:
Closes #406

### DIFF
--- a/src/isar/state_machine/state_machine.py
+++ b/src/isar/state_machine/state_machine.py
@@ -265,9 +265,15 @@ class StateMachine(object):
         fail_statuses: List[TaskStatus] = [
             TaskStatus.Cancelled,
             TaskStatus.Failed,
-            TaskStatus.PartiallySuccessful,
         ]
-        if any(task.status in fail_statuses for task in self.current_mission.tasks):
+        partially_fail_statuses = fail_statuses + [TaskStatus.PartiallySuccessful]
+
+        if all(task.status in fail_statuses for task in self.current_mission.tasks):
+            self.current_mission.status = MissionStatus.Failed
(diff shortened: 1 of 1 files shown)


src/isar/state_machine/state_machine.py
```diff
@@ -301,7 +307,7 @@ def _stop(self) -> None:
 
     def _initiate_step_failed(self) -> None:
         self.current_step.status = StepStatus.Failed
-        self.current_mission.status = MissionStatus.Failed
+        self.current_mission.status = MissionStatus.PartiallySuccessful
```
GodVenn, line 310: Should we set the missions status here or only in the `_mission_finished()` method?
Maybe it made sense when it failed because it would be aborted, but setting it to partially successful implies it will continue, so maybe setting the final status doesn't make sense anymore.

Christdej: Will fix

GodVenn (approved):
LGTM

PR #410, opened 2023-02-23
Make robot status dependent on state machine state

Eddasol:
To fix bug where state machine is in monitor and has a mission but robotStatus is still set as available

### DIFF
--- a/src/isar/services/service_connections/mqtt/robot_status_publisher.py
+++ b/src/isar/services/service_connections/mqtt/robot_status_publisher.py
@@ -5,6 +5,7 @@ from queue import Queue
 from threading import Thread
 
 from isar.config.settings import settings
+from isar.state_machine.states_enum import States
 from isar.state_machine.state_machine import StateMachine
 from robot_interface.models.mission.status import RobotStatus
 from robot_interface.robot_interface import RobotInterface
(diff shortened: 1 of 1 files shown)


GodVenn:
I see this fix as a fix of a symptom instead of a cause, would there be another way where we instead keep the robot status updated based on the state machine, instead of having to check both when publishing it?

aestene:
The robot status is currently gathered from the robot itself in the isar-robot package. At this point there is no information about the state machine available. Thus a conclusion here whether the robot is available, busy or offline is not evidence enough to say that the robot may be reached through ISAR. Let's say ISAR is in a mission and had just completed the previous step. At this point the robot is available but ISAR is busy because it is running a mission but a new step just hasn't been scheduled yet. 

Alternatively, we could have the robot status as a separate field in the State Machine object which is updated regularly as the state machine progresses but I'm not certain it would be a better implementation.

aestene:
In some ways this is a compounded status because what we want to know really is 
1. Am I able to schedule a mission to the robot through ISAR?
2. Is ISAR able to reach the robot or is it offline?

aestene (approved):
Await discussion with @GodVenn before merging.

GodVenn (approved):
Ah I see, then this approach makes sense, I agree

PR #411, opened 2023-02-23
Publish correct task status if partially successful

aestene:


src/isar/models/mission/mission.py
```diff
@@ -56,18 +53,29 @@ def is_finished(self) -> bool:
...
+                return
+
+            elif (step.status is StepStatus.Failed) and isinstance(
+                step, InspectionStep
+            ):
+                self.status = TaskStatus.PartiallySuccessful
+                continue
+
+            elif step.status is StepStatus.Successful:
+                continue
```
GodVenn, line 71: I agree with this functionality, but it's a bit hard to read, maybe we could change it to be "queries" on the array like
if any(blabla in self.steps) ?

GodVenn (approved):
But LGTM!

PR #425, opened 2023-03-07
Revert change to metadata type

aestene:


src/isar/config/settings.py
```diff
@@ -101,7 +102,7 @@ class Settings(BaseSettings):
     AZURE_TENANT_ID: str = Field(default="3aa4a235-b6e2-48d5-9195-7fcf05b459b0")
 
     # Client ID for the API client
-    APP_CLIENT_ID: str = Field(default="68cca82d-84e7-495c-96b4-4c32509f2a46")
+    APP_CLIENT_ID: str = Field(default="fd384acd-5c1b-4c44-a1ac-d41d720ed0fe")
```
aestene, line 105: Accidentally added to the commit but it's a change that should be made so I have kept it.

eivindsjovold (approved):
LGTM

PR #428, opened 2023-03-07
Fail ISAR if any sub-thread fails.

GodVenn:
The previous method would join the first thread in threads and then block for that thread. 
If that thread failed, then it would simply join the next thread, and if another thread failed, it would do nothing.


aestene:
Punctuation in commit message

aestene (approved):
Fix commit message but otherwise LGTM

PR #429, opened 2023-03-09
Allow for input Ids

GodVenn:
Allow for optional ID input to ISAR for mission, tasks and step.
This will allow the user of ISAR (e.g. Flotilla) to track tasks and step if desired.


src/isar/apis/models/start_mission_definition.py
```diff
@@ -25,6 +26,7 @@ class InspectionTypes(str, Enum):
 
 
 class StartMissionInspectionDefinition(BaseModel):
+    id: Optional[str]
```
Christdej, line 29: I think `Optional` parameters should be after all the mandatory parameters

src/isar/apis/models/start_mission_definition.py
```diff
@@ -33,12 +35,14 @@ class StartMissionInspectionDefinition(BaseModel):
 
 
 class StartMissionTaskDefinition(BaseModel):
+    id: Optional[str]
```
Christdej, line 38: I think `Optional` parameters should be after all the mandatory parameters

src/isar/apis/models/start_mission_definition.py
```diff
@@ -33,12 +35,14 @@ class StartMissionInspectionDefinition(BaseModel):
...
     pose: InputPose
     tag: Optional[str]
     inspections: List[StartMissionInspectionDefinition]
 
 
 class StartMissionDefinition(BaseModel):
+    id: Optional[str]
```
Christdej, line 45: I think `Optional` parameters should be after all the mandatory parameters

src/isar/models/mission/mission.py
```diff
@@ -22,7 +23,7 @@ class Task:
     steps: List[STEPS]
     status: TaskStatus = field(default=TaskStatus.NotStarted, init=False)
     tag_id: Optional[str] = field(default=None)
-    id: UUID = field(default_factory=uuid4, init=False)
+    id: Union[UUID, str] = field(default_factory=uuid4, init=True)
```
Christdej, line 26:
```suggestion
    id: str = field(default_factory=uuid4, init=True)
```
Should we change to only use str? I think Union makes for confusion and I dont see the point. Imo we just change the UUID to a str as soon as it is created, then keep it str.

src/robot_interface/models/mission/step.py
```diff
@@ -15,7 +15,7 @@ class Step:
     Base class for all steps in a mission.
     """
 
-    id: UUID = field(default_factory=uuid4, init=False)
+    id: Union[UUID, str] = field(default_factory=uuid4, init=False)
```
Christdej, line 18: Should we change to only use str? I think Union makes for confusion and I dont see the point. Imo we just change the UUID to a str as soon as it is created, then keep it str

src/robot_interface/telemetry/payloads.py
```diff
@@ -36,8 +36,8 @@ class RobotStatusPayload:
     robot_status: RobotStatus
     current_isar_state: State
     current_mission_id: Union[UUID, int, str, None]
-    current_task_id: UUID
-    current_step_id: UUID
+    current_task_id: Union[UUID, str]
+    current_step_id: Union[UUID, str]
```
Christdej, line 40: Should we change to only use str? I think Union makes for confusion and I dont see the point. Imo we just change the UUID to a str as soon as it is created, then keep it str

Christdej (changes requested):

src/isar/models/mission/mission.py
GodVenn: Agree

src/robot_interface/telemetry/payloads.py
GodVenn: Agree

src/isar/apis/models/start_mission_definition.py
```diff
@@ -57,17 +60,31 @@ def to_isar_mission(mission_definition: StartMissionDefinition) -> Mission:
...
                 )
                 for inspection in task.inspections
             ]
         except ValueError as e:
             raise MissionPlannerError(f"Failed to create task: {str(e)}")
+
+        if contains_duplicate_ids(items=inspection_steps):
+            raise MissionPlannerError(
+                f"Failed to create task: Duplicate step IDs are not allowed"
+            )
```
aestene, line 73: Do we allow duplicate ideas in steps from different tasks?

src/isar/apis/models/start_mission_definition.py
GodVenn: fixed

src/isar/apis/models/start_mission_definition.py
GodVenn: fixed

src/isar/apis/models/start_mission_definition.py
GodVenn: fixed

src/isar/apis/models/start_mission_definition.py
GodVenn: Yes, currently, but that was not intended, good catch.
Will fix

GodVenn: Fixed

Christdej (approved):
LGTM

PR #433, opened 2023-03-17
Add RecordAudio inspection step

GodVenn:
Add record audio step because this is supported in Echo it makes sense to support it in ISAR


phhuseboe (approved):
LGTM

PR #435, opened 2023-03-21
State machine

Christdej:
Wait for #423 

Add config setting to run full mission instead of stepwise missions
Renaming of states to prepare for initiate_mission functionallity


aestene (approved):
LGTM

PR #437, opened 2023-03-23
Add available decks

Eddasol:


src/isar/config/settings.py
```diff
@@ -213,6 +214,32 @@ class Settings(BaseSettings):
...
+                    position=Position(x=2, y=2, z=2, frame=Frame(name="asset")),
+                    orientation=Orientation(
+                        x=0, y=0, z=0, w=1, frame=Frame(name="asset")
+                    ),
+                    frame=Frame(name="asset"),
+                ),
+            ),
+        ]
+    )
+
```
eivindsjovold, line 242: I don't know the best practise here but maybe the env variable should be a json that is constructed to an object in code?

Eddasol:
Decided to move assetDeck to fleet manager and not in ISAR

PR #444, opened 2023-03-25
Support full mission

aestene:


Christdej (changes requested):

src/isar/state_machine/state_machine.py
```diff
@@ -458,17 +496,17 @@ def publish_mission_status(self) -> None:
             retain=False,
         )
 
-    def publish_task_status(self) -> None:
-        """Publishes the current task status to the MQTT Broker"""
+    def publish_task_status(self, task: Task) -> None:
```
Afonso-2403, line 499: Should task have a default value so that it can be an optional parameter?

src/isar/state_machine/state_machine.py
```diff
@@ -480,8 +518,8 @@ def publish_task_status(self) -> None:
             retain=False,
         )
 
-    def publish_step_status(self) -> None:
-        """Publishes the current step status to the MQTT Broker"""
+    def publish_step_status(self, step: Step) -> None:
```
Afonso-2403, line 521: Should step have a default value so that it can be an optional parameter?

src/isar/state_machine/state_machine.py
aestene: I don't think so because we will always be publishing task in relation to the executing task. This gives the opportunity to also publish for other tasks than the currently executing one.

src/isar/state_machine/state_machine.py
aestene: Same reason as for task.

Christdej (changes requested):

PR #445, opened 2023-03-27
Fix tests and setup.py

### LINKED ISSUE #414: Replace mock library
We use external "mock" library for some tests when we should use unittest.mock.  Remove mock from setup.py and replace mock with unittest.mock

No change

Improves security by reducing 3rd party library

UsamaEquinorAFK:
Remove json log matches as json codes are already matched. Should close #414

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,6 @@ setup(
         "dev": [
             "black",
             "flake8",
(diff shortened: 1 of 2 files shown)


aestene (approved):
LGTM

PR #449, opened 2023-03-28
Add test for mission definition

tsundvoll:
This test also serves as documentation for how a mission dict / json is transformed into a `Mission` object


Christdej (approved):
GJ

PR #451, opened 2023-03-29
Pass message with robot exception

aestene:
Part of https://github.com/equinor/flotilla/issues/682

Closes https://github.com/equinor/flotilla/issues/682

### DIFF
--- a/src/isar/state_machine/state_machine.py
+++ b/src/isar/state_machine/state_machine.py
@@ -30,6 +30,7 @@ from isar.state_machine.states import (
     Stop,
 )
 from isar.state_machine.states_enum import States
+from robot_interface.models.exceptions.robot_exceptions import ErrorDescription
 from robot_interface.models.initialize.initialize_params import InitializeParams
 from robot_interface.models.mission.mission import Mission
 from robot_interface.models.mission.status import MissionStatus, StepStatus, TaskStatus
@@ -218,24 +219,7 @@ class StateMachine(object):
     #################################################################################
     # Transition Callbacks
     def _initialization_successful(self) -> None:
-        self.queues.start_mission.output.put(True)
-        self.logger.info(
-            f"Initialization successful. Starting new mission: {self.current_mission.id}"
-        )
-        self.log_step_overview(mission=self.current_mission)
-
-        # This is a workaround to enable the Flotilla repository to write the mission to
-        # its database before the publishing from ISAR starts. This is not a permanent
-        # solution and should be further addressed in the following issue.
-        # https://github.com/equinor/flotilla/issues/226
-        time.sleep(2)
-
-        self.current_mission.status = MissionStatus.InProgress
-        self.publish_mission_status()
-        self.current_task = self.task_selector.next_task()
-        self.current_task.status = TaskStatus.InProgress
-        self.publish_task_status(task=self.current_task)
-        self.update_current_step()
+        return
 
     def _initialization_failed(self) -> None:
         self.queues.start_mission.output.put(False)
@@ -265,7 +249,9 @@ class StateMachine(object):
     def _resume(self) -> None:
         self.logger.info(f"Resuming mission: {self.current_mission.id}")
         self.current_mission.status = MissionStatus.InProgress
+        self.current_mission.error_description = None
         self.current_task.status = TaskStatus.InProgress
+
         self.publish_mission_status()
         self.publish_task_status(task=self.current_task)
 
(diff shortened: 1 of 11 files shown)


src/isar/state_machine/state_machine.py
```diff
@@ -257,11 +259,20 @@ def _mission_finished(self) -> None:
         partially_fail_statuses = fail_statuses + [TaskStatus.PartiallySuccessful]
 
         if all(task.status in fail_statuses for task in self.current_mission.tasks):
+            self.current_mission.user_message = UserMessage(
+                full_text=f"The mission failed because all tasks in the mission failed",
```
GodVenn, line 263: Should we add the reasons from the failing tasks here (as done for steps in the task messages)?

src/isar/state_machine/state_machine.py
```diff
@@ -257,11 +259,20 @@ def _mission_finished(self) -> None:
...
+                short_text="all tasks in the mission failed",
+            )
             self.current_mission.status = MissionStatus.Failed
         elif any(
             task.status in partially_fail_statuses
             for task in self.current_mission.tasks
         ):
+            self.current_mission.user_message = UserMessage(
+                full_text="The mission was partially successful because one or more "
+                "tasks failed",
```
GodVenn, line 273: Should we add the reasons from the failing tasks here (as done for steps in the task messages)?

GodVenn (commented):
A great improvement!
What do you think about renaming to "status_reason_message" or something like that?
That way the name describes what it is, instead of what we intend to use it for?

src/isar/state_machine/state_machine.py
```diff
@@ -284,7 +267,24 @@ def _mission_finished(self) -> None:
...
+        self.logger.info(
+            f"Initialization successful. Starting new mission: {self.current_mission.id}"
+        )
+        self.log_step_overview(mission=self.current_mission)
+
+        # This is a workaround to enable the Flotilla repository to write the mission to
+        # its database before the publishing from ISAR starts. This is not a permanent
+        # solution and should be further addressed in the following issue.
+        # https://github.com/equinor/flotilla/issues/226
+        time.sleep(2)
```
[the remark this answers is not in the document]
aestene: Would like to discuss this a bit. As it is part of https://github.com/equinor/flotilla/issues/226 we should solve it there so this PR doesn't wait for it.

tests/isar/state_machine/test_state_machine.py
```diff
@@ -267,8 +273,16 @@ def test_state_machine_with_unsuccsessful_mission_stop(
     )
     actual = state_machine_thread.state_machine.transitions_list
     expected_log = (
-        "Could not communicate request: Reached limit for stop attempts. "
-        "Cancelled mission and transitioned to idle."
+        "Be aware that the robot may still be "
+        "moving even though a stop has been attempted"
     )
     assert expected_log in caplog.text
```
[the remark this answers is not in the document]
aestene: As just the transitions list is not sufficient for that.

src/isar/state_machine/state_machine.py
aestene: See above

src/isar/state_machine/state_machine.py
GodVenn: I'm assuming 'step' here means 'task'.
I agree that this would be nice functionality (Checking each task to see why it failed), and probably a good way to go.
Maybe we should add error codes to our error system and the StatusReason for mission can then say "Mission failed because all tasks failed (errorCode1, errorCode2 etc..)"

src/isar/state_machine/state_machine.py
GodVenn: If we do the above, I think we should remove this description altogether, it's just describing the status without providing any information

tests/isar/state_machine/test_state_machine.py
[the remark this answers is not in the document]
GodVenn: One could argue that it's not testing that, it's just testing that the log says a specific message.
There must be some other way to test this? But if the transitionlist is not enough then it should be a different PR

src/isar/state_machine/state_machine.py
```diff
@@ -333,6 +340,10 @@ def _mission_paused(self) -> None:
         self.current_mission.status = MissionStatus.Paused
         self.current_task.status = TaskStatus.Paused
         self.current_step.status = StepStatus.NotStarted
+        self.current_mission.error_description = ErrorDescription(
+            full_text="The mission was paused manually by an operator",
+            short_text="an operator paused the mission manually",
+        )
```
GodVenn, line 346: This isn't really an error and doesn't need a description I would argue

GodVenn (changes requested):

src/isar/state_machine/state_machine.py
aestene: It's not an error but it does deserve a description in my opinion. Then it is down to the naming of `ErrorDescription` again.

GodVenn (approved):
LGTM

PR #457, opened 2023-04-19
Use exceptions as message carriers for errors

aestene:


tsundvoll:
Can we add a test with just one case that verifies that the user message is present?

src/isar/state_machine/state_machine.py
```diff
@@ -249,7 +249,7 @@ def _off(self) -> None:
     def _resume(self) -> None:
         self.logger.info(f"Resuming mission: {self.current_mission.id}")
         self.current_mission.status = MissionStatus.InProgress
-        self.current_mission.error_description = None
+        self.current_mission.user_message = None
```
tsundvoll, line 252: I like this re-naming :+1:

src/robot_interface/models/exceptions/robot_exceptions.py
```diff
@@ -1,36 +1,96 @@
...
+    RobotMapConfigurationException: str = "robot_map_configuration_exception"
+    RobotMapTransformException: str = "robot_map_transform_exception"
+    RobotUnknownErrorException: str = "robot_unknown_error_exception"
 
 
 @dataclass
-class ErrorDescription:
-    full_text: str
-    short_text: str
+class UserMessage:
```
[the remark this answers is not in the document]
tsundvoll: Agree, and both contents is related to error as well (error_reason and error_message)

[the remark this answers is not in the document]
aestene: Done

aestene:
After discussing the exceptions with @GodVenn new exceptions have been designed and the new suggestions for implementation is ready.

src/isar/state_machine/state_machine.py
```diff
@@ -263,7 +247,9 @@ def _off(self) -> None:
     def _resume(self) -> None:
         self.logger.info(f"Resuming mission: {self.current_mission.id}")
         self.current_mission.status = MissionStatus.InProgress
+        self.current_mission.error_message = None
```
Christdej, line 250: Will this not erase error messages if you pause a mission with error and resume it? That is maybe the point

src/isar/state_machine/states/initialize.py
```diff
@@ -47,11 +51,18 @@ def _run(self) -> None:
...
             try:
                 self.initialize_thread.get_output()
+
             except ThreadedRequestNotFinishedError:
                 time.sleep(self.state_machine.sleep_time)
                 continue
-            except RobotException as e:
-                self.logger.error(f"Initialization of robot failed. Error: {e}")
+
+            except (RobotInitializeException, RobotException) as e:
```
Christdej, line 59: Might be enough to catch RobotException here, as RobotInitializeException inherits from RobotException

src/isar/state_machine/states/monitor.py
```diff
@@ -110,17 +145,12 @@ def _queue_inspections_for_upload(self, current_step: InspectionStep) -> None:
...
-                    f"Error getting inspections for step "
-                    f"{str(current_step.id)[:8]}: {e}"
-                )
-            else:
-                self.logger.error(
-                    f"Error getting inspections for mission "
-                    f"{str(self.state_machine.current_mission.id)[:8]}: {e}"
-                )
+
+        except (RobotRetrieveInspectionException, RobotException) as e:
```
Christdej, line 149: Might be enough to except RobotException as RobotRetrieveInspectionException inherits from RobotException

src/isar/state_machine/states/stop.py
```diff
@@ -48,14 +52,22 @@ def _run(self) -> None:
                 time.sleep(self.state_machine.sleep_time)
                 continue
 
-            except RobotException:
+            except (RobotActionException, RobotException) as e:
```
Christdej, line 55: See above comments

src/robot_interface/models/exceptions/__init__.py
```diff
@@ -1,7 +0,0 @@
-from .robot_exceptions import (
-    RobotCommunicationException,
-    RobotException,
-    RobotInfeasibleMissionException,
-    RobotInfeasibleStepException,
-    RobotInvalidResponseException,
-)
```
Christdej, line 7: Why remove these? Circular dependency exception?

Christdej (changes requested):

src/isar/services/service_connections/mqtt/robot_status_publisher.py
```diff
@@ -86,8 +90,12 @@ def run(self) -> None:
         while True:
             try:
                 self.robot_status = self.robot.robot_status()
-            except RobotCommunicationException:
-                self.logger.warning(
-                    "Failed to get robot status due to a communication exception"
+            except (
+                RobotCommunicationException,
+                RobotAPIException,
+                RobotException,
```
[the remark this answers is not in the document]
aestene: Yes, however I wanted to be specific in showing which exceptions are "expected" for this interface function. I understand that it may be considered bloat, but the interface functions have been documented with specific exceptions (these three) and I wanted to be clear that they are here. I also want to catch any other `RobotExceptions` which might be raised and in this case the behavior for all cases is the same.

src/isar/state_machine/state_machine.py
aestene: From my point of view I think it makes sense that the error message should be removed if you resume the mission. Then if there are problems related to the whole mission again it will reappear beacuse the same error happens again. Note that individual errors for tasks and steps will not be affected.

src/isar/state_machine/states/initialize.py
aestene: Same argument as above.

src/isar/state_machine/states/monitor.py
aestene: Same argument as above.

src/isar/state_machine/states/stop.py
aestene: Same argument as above.

src/robot_interface/models/exceptions/__init__.py
aestene: Yes. 

Also, the pattern which is adding these into `__init__.py` is not friendly towards refactoring activities. It causes some extra import statements but that is handled automatically through packages. So after a discussion with @tsundvoll we decided to remove this.

src/isar/services/service_connections/mqtt/robot_status_publisher.py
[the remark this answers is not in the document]
Christdej: Fine by me

src/isar/state_machine/states/initialize.py
Christdej: Fine by me

src/isar/state_machine/states/monitor.py
Christdej: Fine by me

src/isar/state_machine/states/stop.py
Christdej: Fine by me

src/robot_interface/models/exceptions/__init__.py
Christdej: I agreed that having re-exports in __init__ is not always so refactor-friendly. Really hate to do
`from .../exceptions/exceptions import Exception`, but I will allow it.

Christdej (approved):
LGTM

src/robot_interface/models/exceptions/__init__.py
aestene: If you have an IDE that handles it for you you won't hate it as much as you don't look at it :D

src/isar/state_machine/state_machine.py
```diff
@@ -294,7 +285,25 @@ def _mission_finished(self) -> None:
...
+            f"Initialization successful. Starting new mission: "
+            f"{self.current_mission.id}"
+        )
+        self.log_step_overview(mission=self.current_mission)
+
+        # This is a workaround to enable the Flotilla repository to write the mission to
+        # its database before the publishing from ISAR starts. This is not a permanent
+        # solution and should be further addressed in the following issue.
+        # https://github.com/equinor/flotilla/issues/226
+        time.sleep(2)
```
Christdej, line 299:
```suggestion
        time.sleep(3)
```

PR #467, opened 2023-05-23
Move deepcopy out of ThreadRequest to prevent race conditions

### LINKED ISSUE #466: Mission can be null when uploading final mission inspection
**Describe the bug**
When uploading inspections in several tasks the last mission can be reset before the last upload occurs casuing a null reference.

**To Reproduce**
Run several inspection missions using isar-robot.

**Expected behavior**
All missions should succeed.

**Screenshots**
image

andchiind:
Closes #466

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ setup(
         "backoff",
         "click",
         "dacite",
-        "fastapi-azure-auth",
+        "fastapi-azure-auth==4.0.0",
         "fastapi",
         "injector",
         "opencensus-ext-logging",
--- a/src/isar/state_machine/state_machine.py
+++ b/src/isar/state_machine/state_machine.py
@@ -428,6 +428,7 @@ class StateMachine(object):
         self.publish_state()
 
     def reset_state_machine(self) -> None:
+        self.logger.info("Finalising state machine")
         self.stopped = False
         self.current_step = None
         self.current_task = None
--- a/src/isar/state_machine/state_machine.py
+++ b/src/isar/state_machine/state_machine.py
@@ -428,7 +428,7 @@ class StateMachine(object):
         self.publish_state()
 
     def reset_state_machine(self) -> None:
-        self.logger.info("Finalising state machine")
+        self.logger.info("Resetting state machine")
         self.stopped = False
         self.current_step = None
         self.current_task = None
--- a/src/isar/storage/uploader.py
+++ b/src/isar/storage/uploader.py
@@ -85,6 +85,9 @@ class Uploader:
 
                 inspection, mission = self.upload_queue.get(timeout=1)
 
+                if mission == None:
+                    raise Exception("Mission cannot be null")
+
                 # If new item from thread queue, add one per handler to internal queue:
                 for storage_handler in self.storage_handlers:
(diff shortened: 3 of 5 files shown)


setup.py
```diff
@@ -33,7 +33,7 @@
         "backoff",
         "click",
         "dacite",
-        "fastapi-azure-auth",
+        "fastapi-azure-auth==4.0.0",
```
aestene, line 36: Why was it required to lock this version?

src/isar/state_machine/state_machine.py
```diff
@@ -428,6 +428,7 @@ def update_state(self):
         self.publish_state()
 
     def reset_state_machine(self) -> None:
+        self.logger.info("Finalising state machine")
```
aestene, line 431:
```suggestion
        self.logger.info("Resetting state machine")
```

src/isar/storage/uploader.py
```diff
@@ -85,6 +85,9 @@ def run(self) -> None:
 
                 inspection, mission = self.upload_queue.get(timeout=1)
 
+                if mission == None:
```
aestene, line 88:
```suggestion
                if not mission:
```

src/isar/storage/uploader.py
```diff
@@ -85,6 +85,9 @@ def run(self) -> None:
...
+                    raise Exception("Mission cannot be null")
```
aestene, line 89: This exception will not be caught anywhere as far as I can see which would lead to the Uploader thread crashing and thus the rest of the application crashing. 

As we are doing communication internally within the application it might not be neccessary with this None check at all and I would suggest removing it.

aestene (changes requested):
See comments

setup.py
Christdej: New version of this package had breaking changes, causing isar to crash when launching

src/isar/storage/uploader.py
Christdej: If the mission is None it will crash if we raise and Exception or not, so I guess we change from raising Exception to not trying to upload?

andchiind: Is it not negative to use a continue there without notifying further upstream? As the user in flotilla for example might think the upload has been performed.

Christdej: We need to log something yes, Im thinking perhaps log an error or a warning

aestene: My point is that this is internal communication for the application and if designed properly the mission should never be `None`. It was `None` now due to a bug, not because of some input that was wrong. My suggestion then is to remove the check for `null` entirely because a failure here implies that we have a bug somewhere else, not that something happened which might happen once in awhile.

aestene (changes requested):
Resolve the last comment with the Exception together with @Christdej and then it's good from me.

src/isar/storage/uploader.py
aestene: If you decide that the input should be checked for `None` then make sure it's also caught in the `Uploader` thread such that it won't crash. In that case you should of course log something.

aestene (approved):

PR #470, opened 2023-05-31
Remove unused stid service

anetteu:
Open to close if there is a good reason the keep this in ISAR. As far as I can see, this is dead code and should be cleaned up.


Christdej (approved):
LGTM

aestene (approved):
LGTM

PR #477, opened 2023-06-14
Prevent flotilla from starting several missions

Eddasol:


mrica-equinor (approved):
Tested it and LGTM :)

PR #483, opened 2023-08-09
Split status publisher into heartbeat and status event

aestene:
Implemented as part of https://github.com/equinor/flotilla/issues/954


Afonso-2403 (approved):
lgtm

PR #488, opened 2023-09-20
Draw state diagram using Mermaid

tsundvoll:


tsundvoll:
Closing this PR and keeping the old solution, since that can be autogenerated from the code.

PR #490, opened 2023-10-04
Fix Key Error on analysis type and unlock packages

### LINKED ISSUE #481: Bump pydantic to version 2.x
Pydantic is currently frozen to the latest version 1. Version 2 har some breaking changes, but we would like to bump it still.

Some import structures are changed in pydantic v2.

In general, latest version of dependencies improve the overall threat model by reducing the number of vulnerabilities.

Christdej:
Closes #481 

Changes done according to:
https://docs.pydantic.dev/dev-v2/migration/#validator-and-root_validator-are-deprecated


src/isar/apis/models/models.py
```diff
@@ -11,7 +11,7 @@ class StepResponse(BaseModel):
 
 class TaskResponse(BaseModel):
     id: str
-    tag_id: Optional[str]
+    tag_id: Optional[str] = None
```
aestene, line 14: Why has this been added and have you properly tested this one?

The reason I ask is that for dataclasses atleast it is not supported to have a default value between fields that do not have one. It is probably different for pydantic but just checking to know you have thought about it.

aestene (approved):

PR #491, opened 2023-10-06
Merge ObstacleStatus and RobotStatus

UsamaEquinorAFK:


main.py
```diff
@@ -49,15 +49,16 @@
         target=uploader.run, name="ISAR Uploader", daemon=True
     )
     threads.append(uploader_thread)
-
+    print(settings.MQTT_ENABLED)
```
oysand, line 52: Remember to remove the print

src/isar/services/service_connections/mqtt/robot_status_publisher.py
```diff
@@ -37,6 +37,8 @@ def _get_combined_robot_status(
             return RobotStatus.Offline
         elif current_state == States.Idle and robot_status == RobotStatus.Available:
             return RobotStatus.Available
+        elif current_state != States.Idle or robot_status == RobotStatus.Stuck:
```
oysand, line 40: Should this not be and instead of or?

oysand: Think we only want one catch all on the bottom, but keeping the specific once with and

oysand (changes requested):
See comments

src/isar/services/service_connections/mqtt/robot_status_publisher.py
UsamaEquinorAFK: I think we should decide this regardless of current state and make a decision based on robot_status only.

main.py
```diff
@@ -49,15 +49,15 @@
...
     if settings.MQTT_ENABLED:
+        print("Entered MQTT Enabled block")
         mqtt_client: MqttClient = MqttClient(mqtt_queue=queues.mqtt_queue)
 
         mqtt_thread: Thread = Thread(
             target=mqtt_client.run, name="ISAR MQTT Client", daemon=True
         )
         threads.append(mqtt_thread)
-
+        print("Seeing up robot status publisher now")
```
andchiind, line 60: Is this meant to be "Setting"?

UsamaEquinorAFK: should not be there

PR #492, opened 2023-10-13
Generate steps based on type of task

aestene:
Python version has been raised to 3.10 as `match` is only supported in 3.10 and above. As good a time as any to make the change.

### DIFF
--- a/src/isar/apis/models/start_mission_definition.py
+++ b/src/isar/apis/models/start_mission_definition.py
@@ -10,9 +10,11 @@ from isar.config.settings import settings
 from isar.mission_planner.mission_planner_interface import MissionPlannerError
 from robot_interface.models.mission.mission import Mission
 from robot_interface.models.mission.step import (
-    STEPS,
+    DockingProcedure,
     DriveToPose,
+    Localize,
     RecordAudio,
+    STEPS,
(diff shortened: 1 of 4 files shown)


tsundvoll:
> Python version has been raised to 3.10 as `match` is only supported in 3.10 and above. As good a time as any to make the change.

I agree to this. Since it might interfere with the usage of some of our users, we should consider increasing the minor version to 1.17.x at the next release.

src/isar/apis/models/start_mission_definition.py
```diff
@@ -113,6 +92,64 @@ def to_isar_mission(mission_definition: StartMissionDefinition) -> Mission:
...
+        raise MissionPlannerError(
+            f"Failed to create as there were duplicate IDs which is not allowed "
+            f"({duplicate_ids})"
+        )
+
+
+def generate_steps(task) -> List[STEPS]:
+    try:
+        steps: List[STEPS] = []
+
```
tsundvoll, line 107: I would argue it is better to have this outside the `try catch`, to minimize the code inside it

aestene: Done

tsundvoll (approved):

PR #496, opened 2023-10-25
Add missing parameter for /info/robot-settings endpoint

tsundvoll:
as well as adding missing test for the info endpoint (this new test is failing without the fix)


Christdej (approved):
LGTM

PR #500, opened 2023-11-06
Add test for StartMissionDefinition

tsundvoll:
since we could have better test coverage for the `to_isar_mission(mission_definition)` function

### DIFF
--- a/tests/isar/models/test_start_mission_definition.py
+++ b/tests/isar/models/test_start_mission_definition.py
@@ -1,8 +1,16 @@
+import json
+import os
 from typing import List, Union
 
 import pytest
+from alitra import Frame, Orientation, Pose, Position
 
-from isar.apis.models.start_mission_definition import get_duplicate_ids
+from isar.apis.models.start_mission_definition import (
+    StartMissionDefinition,
+    get_duplicate_ids,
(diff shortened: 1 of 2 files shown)


tests/isar/models/test_start_mission_definition.py
```diff
@@ -51,3 +59,36 @@ def test_duplicate_id_check(
...
+    assert generated_mission.id == "generated_mission_id"
+    assert generated_mission.name == "my-mission"
+    assert len(generated_mission.tasks) == 1
+
+    task = generated_mission.tasks[0]
+    assert task.id == "generated_task_id"
+    assert task.tag_id == "MY-TAG-123"
+
+    first_step = task.steps[0]
+    # assert first_step.id == "generated_inspection_id"
```
andchiind, line 82: Do these fail?

andchiind (approved):
LGTM

tests/isar/models/test_start_mission_definition.py
tsundvoll: Yes, line 82 would fail, therefore it is commented out for now. It might be desired behavior, since the .id attribute will be set by the constructor.

PR #509, opened 2024-01-09
Fix gcc error when building Dockerfile

tsundvoll:
changing base image to avoid Pydantic error:
`ModuleNotFoundError: No module named 'pydantic_core._pydantic_core'`


UsamaEquinorAFK (approved):
LGTM 👯

PR #515, opened 2024-02-16
Fix missing argument in mqtt.Client

tsundvoll:
due to new breaking change in paho.mqtt library starting from v2.0


UsamaEquinorAFK (approved):
lgtm

PR #522, opened 2024-02-22
Handle initiate infeasible for full missions

Afonso-2403:
Initate infeasible for full missions
should just fail as there are no next steps
to be tried


andchiind (approved):
LGTM, tested and it works. In the future we may want to find a way to only cancel the infeasible tasks.

PR #527, opened 2024-03-01
Implement pause mission

andchiind:
I will test this more with isar-exr. We might need the interface functions to return a value so that we can check if they were successful. If they were not then we should transition back to the previous state.


andchiind:
See https://github.com/equinor/isar/pull/565

PR #528, opened 2024-03-01
Update published blob storage inspection path

### LINKED ISSUE #525: Publish descriptive MQTT message on upload to Blob Storage
When an inspection result is stored in Blob Storage, a reference to the data should be provided in a message on the "inspection_result" topic over MQTT. It should be possible for another system to later retrieve the inspection result from the blob storage using only the information provided in the message.

When an image or another inspection result is stored in a blob storage, retrieve information about how this can be accessed later. Part of the issue is to figure out technically how this is done with blob storage, whether it is an ID or an url etc.

Make sure this ID / URL / other reference is published over MQTT on the "inspection_result" topic. Something similar is already in place, but I suspect the ID provided is not working.

A good starting point is the `isar/storage/uploader.py` and `isar/storage/blob_storage` modules, where a message is published to the "inspection_result" topic.

Verify that such a message can be received and that the information provided is enough to retrieve the inspection result back from the blob storage.

N/A

sondreo:
Will resolve #525.

### DIFF
--- a/src/isar/storage/blob_storage.py
+++ b/src/isar/storage/blob_storage.py
@@ -55,7 +55,7 @@ class BlobStorage(StorageInterface):
         except Exception as e:
             self.logger.error("An unexpected error occurred while uploading blob")
             raise StorageException from e
-        return blob_properties["etag"]
+        return blob_client.url
 
     def _get_blob_service_client(self) -> BlobServiceClient:
         try:


sondreo:
Changes the inspection data path returned by the upload-function in the `BlobStorage` module. The full url to the blob will be utilized as a topic in the MQTT message published upon upload.

This url can then be utilized to download or stream the blob, for instance as suggested [here](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-python?tabs=managed-identity%2Croles-azure-portal%2Csign-in-azure-cli&pivots=blob-storage-quickstart-scratch#download-blobs).

sondreo:
Closing, as it has been solved by https://github.com/equinor/isar/pull/529

PR #529, opened 2024-03-08
Proposed solution to send inspection result over mqtt

### LINKED ISSUE #525: Publish descriptive MQTT message on upload to Blob Storage

mrica-equinor:
Closes #525

### DIFF
--- a/src/isar/storage/test_download_blob.py
+++ b/src/isar/storage/test_download_blob.py
@@ -1,0 +1,42 @@
+import os
+from isar.config.keyvault.keyvault_service import Keyvault
+from isar.config.settings import settings
+from isar.storage.blob_storage import BlobStorage
+
+
+class TestBlobdownload:
+    def __init__(self):
+        self.keyvault: Keyvault = Keyvault(
+            keyvault_name=settings.KEYVAULT_NAME,
+            client_id=settings.AZURE_CLIENT_ID,
+            client_secret=None,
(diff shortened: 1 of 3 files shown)


src/isar/storage/test_download_blob.py
```diff
@@ -0,0 +1,42 @@
+import os
+from isar.config.keyvault.keyvault_service import Keyvault
+from isar.config.settings import settings
+from isar.storage.blob_storage import BlobStorage
+
+
+class TestBlobdownload:
```
tsundvoll, line 7: Not sure if we want to have this test in the repo. Do we want it to run in CI/CD?

PR #530, opened 2024-03-13
Refer to blob storage in inspection result message

tsundvoll:
when sending over MQTT


betaniat (approved):
change in inspection_path to absolute inspection_path_dict. Currently working.

PR #534, opened 2024-04-02
Send robot capabilities in info event

### LINKED ISSUE #533: Send robot capabilities in RobotInfoPublisher
We should send robot settings.CAPABILITIES in RobotInfoPublisher.

Currently ISAR communication mainly works as a black box for robots where outside systems have little understanding of the implementation of each ISAR instance. This is good for keeping the system agnostic and simple, but it introduces some limitations which I think can be fixed relatively easily. For instance, each robot has a set of "CAPABILITIES" which they support. Using one of these in a request when they are not supported results in a returned error which the outside system could not have predicted without hardcoding their pre-existing understanding of the ISAR instance. Allowing them to get this information after they first detect that a robot is connected to MQTT would fix this issue and make the system more loosely coupled, as the capabilities is a business logic which currently needs to be "hardcoded" elsewhere.

It will not affect it as long as we only give information about the CAPABILITIES field.

andchiind:
Closes #533


oysand (approved):
LGTM

PR #536, opened 2024-04-03
Introduce offline state

oysand:


oysand:
Continued in https://github.com/equinor/isar/pull/549

PR #549, opened 2024-05-02
Offline state refactor robot status

oysand:


src/isar/state_machine/states/idle.py
```diff
@@ -37,4 +48,36 @@ def _run(self) -> None:
                 break
             time.sleep(self.state_machine.sleep_time)
 
+            time_from_last_robot_status_poll = (
+                time.time() - self.last_robot_status_poll_time
```
oysand, line 52: Reset time

src/isar/state_machine/states/idle.py
```diff
@@ -37,4 +48,38 @@ def _run(self) -> None:
...
+            try:
+                robot_status: RobotStatus = self.robot_status_thread.get_output()
+            except ThreadedRequestNotFinishedError:
+                time.sleep(self.state_machine.sleep_time)
+                continue
+
+            except (RobotException,) as e:
+                self.logger.error(
+                    f"Failed to get robot status because: {e.error_description}"
+                )
```
andchiind, line 77: Could we maybe move this to another function in order to shorten the function and make it a bit easier to read?

src/isar/state_machine/states/idle.py
```diff
@@ -37,4 +48,38 @@ def _run(self) -> None:
...
+
+            self.last_robot_status_poll_time = time.time()
+
+            if robot_status == RobotStatus.Offline:
+                transition = self.state_machine.robot_turned_offline  # type: ignore
+                break
+
```
andchiind, line 84: Alternatively we could move all this to another function

src/isar/state_machine/state_machine.py
```diff
@@ -580,25 +585,33 @@ def publish_step_status(self, step: Step) -> None:
...
             retain=False,
         )
 
+    def _current_status(self) -> RobotStatus:
+        if self.current_state == States.Idle:
+            return RobotStatus.Available
+        elif self.current_state == States.Offline:
+            return RobotStatus.Offline
+        else:
+            return RobotStatus.Busy
```
andchiind, line 613: Do we need a blocked state to map to RobotStatus.Blocked?

andchiind: Not needed in this PR, but might be re-added in other PR

src/isar/state_machine/states/idle.py
andchiind: We can standardise this in another PR

src/isar/state_machine/states/idle.py
andchiind: We can standardise this in another PR

PR #565, opened 2024-06-05
New implement pause

mrica-equinor:


src/isar/state_machine/state_machine.py
```diff
@@ -125,8 +125,8 @@ def __init__(
                 {
                     "trigger": "pause",
                     "source": [self.initiate_state, self.monitor_state],
-                    "dest": self.stop_state,
-                    "before": self._pause,
+                    "dest": self.paused_state,
```
andchiind, line 128: This might be a breaking change for step_wise robots

mrica-equinor: We can add a new transition and if it's not stepwise mission call that new transition that doesn't go trough the stop state. Will add a TODO

mrica-equinor: image


image

PR #570, opened 2024-06-06
Ignore localize ability if running full missions

tsundvoll:


andchiind (approved):
This is a temporary fix which will be addressed when we move away from localisation missions

PR #589, opened 2024-07-10
Add robot execption when amending mission

mrica-equinor:


andchiind (commented):
Could we maybe make this error a bit more general? Such as RobotMissionInitialisationError or RobotMissionAPIError or something like that?

PR #590, opened 2024-07-15
Refactor monitor step status reporting

andchiind:


mrica-equinor (approved):
LGTM :) we could try to find a better word for increment_current_step

betaniat (commented):
LGTM :)

PR #602, opened 2024-08-09
Fix stepwise task reporting not reporting successful

### LINKED ISSUE #597: Stepwise step reporting no longer reports steps as successful
**Describe the bug**
Currently ISAR never reports steps as completed when using stepwise missions. It does however report tasks as in progress.

**To Reproduce**
Set ISAR to use stepwise missions for ISAR_robot. Do not change any other settings from default.

**Expected behavior**
It should behave the same as missionwise reporting.

**Screenshots**

andchiind:
Closes #597


mrica-equinor (approved):
LGTM :)

PR #604, opened 2024-08-29
Add configurable documentation variable

Eddasol:


src/isar/config/settings.py
```diff
@@ -190,6 +190,16 @@ def __init__(self) -> None:
     # Serial number of the robot ISAR is connected to
     SERIAL_NUMBER: str = Field(default="0001")
 
+    # Info about robot documentation
+    DOCUMENTATION: List[DocumentInfo] = Field(
+        default=[
+            DocumentInfo(
+                name="Some document",
+                url="Some URL",
```
andchiind, line 198: Do we need these placeholders?

andchiind (approved):
Looks good

src/isar/config/settings.py
Eddasol: Changing default to empty list

PR #633, opened 2024-10-23
Accept inspection data asynchronously

andchiind:


src/isar/script.py
```diff
@@ -103,6 +106,18 @@ def start():
         target=uploader.run, name="ISAR Uploader", daemon=True
     )
     threads.append(uploader_thread)
+
+    if settings.UPLOAD_INSPECTIONS_ASYNC:
+
+        def inspections_callback(inspection: Inspection):
+            message: Tuple[Inspection, Mission] = (
+                inspection,
+                state_machine.current_mission,
```
andchiind, line 115: Be aware that here we assume that inspection events do not occur very late, ie. that the current mission has not changed between a task completing and its inspection event being received.

andchiind: If this becomes a problem then we could make the callback accept a Mission object, but I would prefer not to require this as it's a bit obtuse.

mrica-equinor (approved):
LGTM :)

PR #634, opened 2024-10-24
Allow stopping missions from idle state

haakonsf:


andchiind (approved):
LGTM, but might want to label this change a breaking change, as it will affect ISAR implementations. We need to make sure they do not crash if they receive the "stop" function call when they don't have a mission running.

haakonsf:
Postponing merging this until isar-taurob test

PR #651, opened 2024-11-28
Get media configs in API endpoint

andchiind:
See https://github.com.mcas.ms/equinor/isar/pull/645 for earlier discussions


src/isar/modules.py
```diff
@@ -47,6 +50,14 @@ def provide_scheduling_controller(
     ) -> SchedulingController:
         return SchedulingController(scheduling_utilities)
 
+    @provider
+    @singleton
+    def provide_robot_controller(
+        self,
+        robot_utilities: RobotUtilities,
+    ) -> SchedulingController:
```
sondreo, line 58:
```suggestion
    ) -> RobotController:
```

src/isar/services/utilities/robot_utilities.py
```diff
@@ -0,0 +1,27 @@
+import logging
+
+from injector import inject
+
+from isar.config.settings import settings
+
+from isar.config.settings import settings
+
```
sondreo, line 8:
```suggestion
```
Remove duplicate import.

src/isar/services/utilities/robot_utilities.py
```diff
@@ -0,0 +1,27 @@
...
+from isar.apis.models.models import MediaConfig
+from robot_interface.robot_interface import RobotInterface
+
+
+class RobotUtilities:
+    """
+    Contains utility functions getting robot information from the API.
```
sondreo, line 15:
```suggestion
    Contains utility functions for getting robot information from the API.
```

src/isar/apis/models/models.py
```diff
@@ -33,6 +35,13 @@ class RobotInfoResponse(BaseModel):
     plant_short_name: str
 
 
+@dataclass
```
sondreo, line 38: Should we consider making the `MediaConfig` response object a Pydantic Model by inheriting from `BaseModel`?

This could alternatively be deferred to a separate improvement issue.

sondreo (changes requested):
Looks good! See some minor suggestions.

src/isar/apis/models/models.py
andchiind: Yes, I think it might be easiest to do it in a separate issue https://github.com.mcas.ms/equinor/isar/issues/655

PR #660, opened 2024-12-03
Let ISAR generate mission.id, task.id and inspection.id

tsundvoll:


src/isar/apis/models/start_mission_definition.py
```diff
@@ -44,188 +43,139 @@ class StartMissionInspectionDefinition(BaseModel):
...
 
 
 class StartMissionDefinition(BaseModel):
     tasks: List[StartMissionTaskDefinition]
-    id: Optional[str] = None
     name: Optional[str] = None
     start_pose: Optional[InputPose] = None
-    dock: Optional[bool] = None
-    undock: Optional[bool] = None
+    dock: Optional[bool] = Field(default=False)
```
oysand, line 60: Check if False as default trigger other behaviour than None

src/isar/state_machine/states/monitor.py
```diff
@@ -13,13 +13,13 @@
...
     ErrorMessage,
+    RobotCommunicationException,
     RobotCommunicationTimeoutException,
     RobotException,
     RobotRetrieveInspectionException,
     RobotTaskStatusException,
-    RobotCommunicationException,
 )
-from robot_interface.models.inspection.inspection import Inspection
+from robot_interface.models.inspection.inspection import InspectionResult
```
oysand, line 22: Rename back to inspection, but now inspections are not created before the result is ready for uploading

oysand (approved):
LGTM

PR #666, opened 2024-12-06
Pass on task zoom value from API

andchiind:


src/robot_interface/models/mission/task.py
```diff
@@ -29,6 +29,12 @@ class TaskTypes(str, Enum):
     DockingProcedure = "docking_procedure"
 
 
+@dataclass
+class ZoomDescription:
+    width: float
+    height: float
```
Eddasol, line 35: Since the values are height and width of inspection object it might be more descriptive to rename to something like ObjectSize/InspectionTargetSize?

eivindsjovold: Agree with @Eddasol, InspectionTargetSize or InspectionTargetProperties

andchiind: I compromised and made it clear that the width and height refer to the width and height of the object, while keeping it in a "zoom" context, since it is not the general width and height of the object, but the width and height from the point of view of the robot when taking a photo. This could otherwise create some confusion imo. We for instance do not provide a depth, so it is not a complete description of the inspection target.

andchiind: But if we end up adding more metadata for the inspection action then we can also generalise this class.

andchiind: We are not also necessarily describing the objects' width and height but just the width and height of the target in the image. For a fence for instance, we are not stating that the provided values are the width and height of the fence.

mrica-equinor (approved):
LGTM :)

PR #672, opened 2024-12-11
Remove unused video stream object

Eddasol:
Dependent on https://github.com/equinor/flotilla/pull/1869

### DIFF
--- a/src/isar/config/settings.py
+++ b/src/isar/config/settings.py
@@ -8,7 +8,7 @@ from pydantic_settings import BaseSettings, SettingsConfigDict
 
 from isar.config import predefined_missions
 from robot_interface.models.robots.robot_model import RobotModel
-from robot_interface.telemetry.payloads import DocumentInfo, VideoStream
+from robot_interface.telemetry.payloads import DocumentInfo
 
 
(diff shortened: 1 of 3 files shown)


src/isar/config/settings.py
```diff
@@ -194,22 +194,6 @@ def __init__(self) -> None:
     # Info about robot documentation
     DOCUMENTATION: List[DocumentInfo] = Field(default=[])
 
-    # Endpoints to reach video streams for the robot
-    VIDEO_STREAMS: List[VideoStream] = Field(
```
sondreo, line 198: This deleted settings variable is also used in [RobotInfoPublisher](https://github.com/equinor/isar/blob/c8ff71cdf9bd409c3aceb59567c877251a15d71d/src/isar/services/service_connections/mqtt/robot_info_publisher.py#L25), and should be removed there as well

sondreo (changes requested):
See small comment, otherwise good 👍

src/isar/config/settings.py
Eddasol: Good catch :)

sondreo (approved):

PR #675, opened 2024-12-12
Introduce battery state to battery payload

Eddasol:


src/robot_interface/telemetry/payloads.py
```diff
@@ -30,6 +31,7 @@ class TelemetryPosePayload(TelemetryPayload):
 @dataclass
 class TelemetryBatteryPayload(TelemetryPayload):
     battery_level: float
+    battery_state: Optional[BatteryState] = None
```
sondreo, line 34: Suggestion: Keep the type required / non-optional, and handle missing values for the payload in the robot implementation. This way it will match the implementation of the fields of the other data classes. The `battery_level` missing telemetry value is handled in the robot implementation. 

@Eddasol any thoughts?

sondreo: As discussed, it will be kept as is.

sondreo (approved):
LGTM! Has been tested locally 👍

PR #678, opened 2024-12-12
Reintroduce isar task id in mission task definition

tsundvoll:
for Flotilla to be able to match its existing task request to the one that is returned from ISAR


oysand (approved):
LGTM

PR #681, opened 2024-12-12
Use payloads for mqtt messages

Eddasol:


src/robot_interface/telemetry/payloads.py
```diff
@@ -88,3 +84,38 @@ class RobotHeartbeatPayload:
...
+    error_description: Optional[str]
+    timestamp: datetime
+
+
+@dataclass
+class InspectionResultPayload:
+    isar_id: str
+    robot_name: str
+    inspection_id: str
+    inspection_path: str | dict
```
Eddasol, line 118:
```suggestion
    inspection_path: Union[str, dict]
```

src/robot_interface/telemetry/payloads.py
```diff
@@ -82,3 +78,38 @@ class RobotHeartbeatPayload:
...
+
+
+@dataclass
+class InspectionResultPayload:
+    isar_id: str
+    robot_name: str
+    inspection_id: str
+    inspection_path: Union[str, dict]
+    installation_code: str
+    analysis_to_be_run: Optional[str]
```
sondreo, line 114: Does this have to be optional?

sondreo (approved):
LGTM! Has been tested locally.

See a minor non-blocking comment.

src/robot_interface/telemetry/payloads.py
Eddasol: Yes. Its being used the send analysis_type that is defined like: 
`analysis_type: Optional[str] = field(default=None, init=False)`

PR #684, opened 2024-12-16
Add ruff workflow to prevent unused imports

Eddasol:
Removing InitializeParams from init file means that all isar-"robot" repos using the line below:
`from robot_interface.models.initialize import InitializeParams
`
will have to change it to:
`from robot_interface.models.initialize.initialize_params import InitializeParams`


tsundvoll (approved):
This is great! Good idea to implement ruff to the linting 👍

PR #689, opened 2025-01-03
Introduce conditional transitions and move status threads to new service

### LINKED ISSUE #687: Move initiate and initialize state to a transition
We currently have an initiate state since we want to check that we can successfully start a mission before moving to monitor. Using conditional transitions, this check can be done in the transition between idle and monitor, removing the need for the inbetween states. This will move us towards the goal of not doing state changes inside states themselves, so that states merely receive events and the transitions perform the changes themselves.

It will simplify the state machine architecture.

N/A

andchiind:
Closes #687 

This PR removes the initialize and initiate states, tidies up transition functions and moves several threads out of the state machine states and into a new robot service. The statemachine communicates with this robot service using the threadsafe queues. The existing ones have been given new names to make it easier to distinguish where they are used. They can be considered events and the functions that listen for them are event listeners. In the future I would like to refactor the different states so that they are more cleanly organised as a set of event handlers (https://github.com/equinor/isar/issues/715). Other future work after this PR would be to also move the stop and pause threads into the robot service (https://github.com/equinor/isar/issues/712), and to split up the different queues into their own subclasses (https://github.com/equinor/isar/issues/720). We will of course also need to update the diagram in the readme (https://github.com/equinor/isar/issues/717).

This issue has also been suggested as a part of the review process https://github.com/equinor/isar/issues/731.

- [ ] A test has been written


src/isar/state_machine/state_machine.py
```diff
@@ -182,13 +160,19 @@ def __init__(
                 {
                     "trigger": "task_finished",
                     "source": self.monitor_state,
-                    "dest": self.initiate_state,
-                    "before": self._task_finished,
+                    "dest": self.monitor_state,
+                    "conditions": self._start_task,
+                },
+                {
```
andchiind, line 166: This transition is basically the "else" which is run if the conditional above returns false. In this case, if it fails to continue the mission (for instance if there are no more tasks) it will go to the idle state and run the "_full_mission_finished" cleanup function.

src/isar/state_machine/state_machine.py
```diff
@@ -328,17 +334,101 @@ def _mission_started(self) -> None:
...
+            self.logger.warning(
+                f"Failed to initiate mission "
+                f"{str(self.current_mission.id)[:8]} because: "
+                f"{e.error_description}"
+            )
+            self._initiate_failed()
+            return False
+        return True
+
+    def _try_start_mission(self) -> bool:
```
andchiind, line 411: This contains the functionality that used to the in the initiate and intialisation states

src/isar/state_machine/state_machine.py
andchiind: The same is true for the transition from idle to idle. This is for when we fail to start a mission.

src/isar/state_machine/transitions/start_mission.py
```diff
@@ -0,0 +1,82 @@
...
+    )
+    state_machine.log_mission_overview(mission=state_machine.current_mission)
+
+    state_machine.current_mission.status = MissionStatus.InProgress
+    state_machine.publish_mission_status()
+    state_machine.current_task = state_machine.task_selector.next_task()
+    state_machine.send_task_status()
+    if state_machine.current_task is None:
+        return False
+    else:
```
oysand, line 34: Can this else be removed and indent the code one level back?

src/isar/robot/robot.py
```diff
@@ -0,0 +1,261 @@
...
+                    and self.start_mission_thread._is_thread_alive()
+                ):
+                    self.start_mission_thread.wait_for_thread()
+                self.start_mission_thread = ThreadedRequest(
+                    request_func=self.run_start_mission_thread,
+                )
+                self.start_mission_thread.start_thread(
+                    start_mission_or_task, name="Robot start mission thread"
+                )
+                # TODO: consider what happens when we get another start_mission while the start_mission_thread is running.
```
oysand, line 252: Remove todo

src/isar/robot/robot.py
```diff
@@ -0,0 +1,261 @@
...
+from robot_interface.models.mission.task import Task
+from robot_interface.models.exceptions.robot_exceptions import (
+    ErrorMessage,
+    RobotException,
+    RobotInfeasibleTaskException,
+    RobotInfeasibleMissionException,
+)
+
+
+class Robot(object):
```
oysand, line 36: Should this be split further so there is a separate class per thread? Similarly to how the telemetry is handled?

src/isar/robot/robot.py
```diff
@@ -0,0 +1,261 @@
...
+                    self.queues.robot_task_status_failed,
+                    failed_task_error,
+                )
+                task_status = TaskStatus.Failed
+            else:
+                self._trigger_event(self.queues.robot_task_status, task_status)
+            self.current_status = task_status
+            time.sleep(settings.FSM_SLEEP_TIME)
+
+    def stop(self) -> None:
```
oysand, line 155: when is this triggered?

src/isar/robot/robot.py
```diff
@@ -0,0 +1,261 @@
...
+            self._trigger_event(
+                self.queues.robot_mission_failed,
+                ErrorMessage(
+                    error_reason=e.error_reason, error_description=e.error_description
+                ),
+            )
+        self.robot_task_status_thread = ThreadedRequest(
+            request_func=self.run_robot_task_status_thread
+        )
+        self.robot_task_status_thread.start_thread(name="Robot task status thread")
```
oysand, line 223: how is this stopped when there is no mission active?

andchiind: The thread exits when it transitions out of monitor, and it is then cleaned up when receiving another mission

src/isar/robot/robot.py
andchiind: It's not used yet, but I might use it before moving the PR out of draft

src/isar/robot/robot.py
andchiind: I will leave this for a future refactor.

src/isar/robot/robot.py
andchiind: Will be removed before moving the PR out of draft

src/isar/robot/robot.py
andchiind: I will change it so that the thread stays alive, but only sleeps while no mission or task is active. This way we avoid the cost of having to start another thread when a new task starts. Alternatively I might make it start when we detect that we enter monitor, then exit when we leave monitor.

src/isar/robot/robot.py
andchiind: It is now used in test cleanup

src/isar/state_machine/state_machine.py
```diff
@@ -390,31 +280,61 @@ def start_mission(self, mission: Mission, initial_pose: Pose):
...
         except queue.Empty:
             return False
 
+    def has_mission_failed(self) -> Optional[ErrorMessage]:
+        try:
+            return self.queues.robot_mission_failed.input.get(block=False)
+        except queue.Empty:
+            return None
+
+    def get_task_failure(self) -> Optional[ErrorMessage]:
```
mrica-equinor, line 320: I find this name confusing - maybe get_task_failure_reason? for a second there i thought you were fetching a failed task. I also was able to quickly figure out what the function actually was doing so

src/isar/state_machine/states/monitor.py
```diff
@@ -61,50 +52,51 @@ def _run(self) -> None:
...
-            ) as e:
-                retry_limit_exceeded: bool = self._check_if_exceeded_retry_limit(e)
-                if retry_limit_exceeded:
-                    self.logger.error(
-                        f"Monitoring task {self.state_machine.current_task.id[:8]} failed "
-                        f"because: {e.error_description}"
-                    )
-                    status = TaskStatus.Failed
+            if not self.state_machine.mission_has_started:
+                if self.state_machine.has_mission_started():
```
mrica-equinor, line 56: suggest to rename mission_has_started to mission_started or something. this bit of code made no sense at first, i get the point but it looks weird

mrica-equinor: mission_ongoing would be good too

src/isar/state_machine/state_machine.py
andchiind: I can maybe call it get_task_failure_event?

src/isar/state_machine/states/monitor.py
andchiind: I have tried to address this by making it more obvious that it is events we are waiting for

src/isar/state_machine/state_machine.py
mrica-equinor: sounds good to me, less confusing at least

src/isar/models/communication/queues/queues.py
```diff
@@ -7,11 +7,26 @@
 
 class Queues:
     def __init__(self) -> None:
-        self.start_mission: QueueIO = QueueIO(input_size=1, output_size=1)
-        self.stop_mission: QueueIO = QueueIO(input_size=1, output_size=1)
-        self.pause_mission: QueueIO = QueueIO(input_size=1, output_size=1)
-        self.resume_mission: QueueIO = QueueIO(input_size=1, output_size=1)
-        self.single_action: QueueIO = QueueIO(input_size=1, output_size=1)
+        self.api_start_mission: QueueIO = QueueIO(input_size=1, output_size=1)
```
oysand, line 10: Could it be worth splitting this into several queues classes so it is clear which threads they facilitate communication between?

andchiind: You mean having the response in a different queue than the request? Currently these are just renamed versions of the already existing queues, but I could make an issue on splitting it up so that the API thread does not listen to the same thread where it sends the request to the state machine.

andchiind: Ok I think I understand what you mean. So that each group of queues should be its own subclass of the queue utility class? I could do that

andchiind: Here you go https://github.com/equinor/isar/issues/720

src/isar/robot/robot.py
```diff
@@ -0,0 +1,90 @@
...
+            and self.robot_status_thread.is_alive()
+        ):
+            self.start_mission_thread.join()
+        self.robot_status_thread = None
+        self.robot_task_status_thread = None
+        self.start_mission_thread = None
+        self.signal_thread_quitting.clear()
+
+    def handle_stop_mission(self):
+        return NotImplementedError()
```
oysand, line 45: The function should not return NotImplementedError. If it is to be implemented in the robot package, then it should call a function in the robot_interface that raises NotImplementedError. Else it should be removed

.github/workflows/pythonpackage.yml
```diff
@@ -50,7 +50,7 @@ jobs:
 
       - name: Run mypy
         run: |
-          mypy --disable-error-code=import-untyped .
+          mypy --disable-error-code=import-untyped --check-untyped-defs .
```
Christdej, line 53: Why add this?

src/isar/robot/robot.py
```diff
@@ -0,0 +1,90 @@
...
+            start_mission_or_task = check_for_event(
+                self.queues.state_machine_start_mission
+            )
+            if start_mission_or_task is not None:
+                if (
+                    self.start_mission_thread is not None
+                    and self.start_mission_thread.is_alive()
+                ):
+                    self.start_mission_thread.join()
+                self.start_mission_thread = RobotStartMissionOrTaskThread(
```
oysand, line 68: Remove start task part of this and only start the mission

src/isar/models/communication/queues/queues.py
```diff
@@ -7,11 +7,26 @@
...
+        self.api_stop_mission: QueueIO = QueueIO(input_size=1, output_size=1)
```
Christdej, line 11: Is there any way for a user of ISAR that have queued a mission, to see that the mission is in the queue? (nothing to do with this line of code, just put the question here)

src/isar/robot/robot.py
```diff
@@ -0,0 +1,90 @@
...
+        )
+        self.robot_status_thread.start()
+
+        while True:
+            if self.signal_thread_quitting.is_set():
+                break
+            start_mission_or_task = check_for_event(
+                self.queues.state_machine_start_mission
+            )
+            if start_mission_or_task is not None:
```
oysand, line 62: can this be rewritten as guard statements?

src/isar/robot/robot.py
```diff
@@ -0,0 +1,90 @@
...
+            if check_for_event(self.queues.state_machine_stop_mission):
+                if (
+                    self.start_mission_thread is not None
+                    and self.start_mission_thread.is_alive()
+                ):
+                    self.start_mission_thread.join()
+                self.handle_stop_mission()
+            if check_for_event(self.queues.state_machine_pause_mission):
+                self.handle_pause_mission()
+            task_id = check_for_event(self.queues.state_machine_get_task_status)
```
oysand, line 84: Change state_machine_get_task_status to state_machine_request_new_task_status_event

src/isar/robot/robot.py
```diff
@@ -0,0 +1,90 @@
...
+                    self.start_mission_thread.join()
+                self.start_mission_thread = RobotStartMissionOrTaskThread(
+                    self.queues,
+                    self.robot,
+                    self.signal_thread_quitting,
+                    start_mission_or_task,
+                )
+                self.start_mission_thread.start()
+            if check_for_event(self.queues.state_machine_stop_mission):
+                if (
```
oysand, line 76: Bake into function called _wait_for_thread_to_finish or something that describes the impact of the if check

src/isar/robot/robot.py
andchiind: I don't really think so tbh, since I don't want to stop the loop if the event does not exist. This will be more readible with this change though: https://github.com/equinor/isar/issues/715

src/isar/robot/robot.py
andchiind: It was related to this issue that they are currently not implemented: https://github.com/equinor/isar/issues/712

src/isar/robot/robot.py
```diff
@@ -0,0 +1,90 @@
...
+                self.handle_stop_mission()
+            if check_for_event(self.queues.state_machine_pause_mission):
+                self.handle_pause_mission()
+            task_id = check_for_event(self.queues.state_machine_get_task_status)
+            if task_id:
+                self.robot_task_status_thread = RobotTaskStatusThread(
+                    self.queues, self.robot, self.signal_thread_quitting, task_id
+                )
+                self.robot_task_status_thread.start()
+        self.logger.info("Exiting robot service main thread")
```
tsundvoll, line 90:
```suggestion
        while True:
            if self.signal_thread_quitting.is_set():
                break
            check_and_handle_start_mission(self.queues)
            check_and_handle_stop_mission(self.queues)
            check_and_handle_pause_mission(self.queues)
            check_and_handle_get_task_status(self.queues)
        self.logger.info("Exiting robot service main thread")
```

And have the handeling happening inside functions like for example this:

```python
def check_and_handle_start_mission(events):
    start_mission = check_for_event(events.state_machine_start_mission)
    if start_mission is not None:
        if (
            self.start_mission_thread is not None
            and self.start_mission_thread.is_alive()
        ):
            self.start_mission_thread.join()
        self.start_mission_thread = RobotStartMissionOrTaskThread(
            self.queues,
            self.robot,
            self.signal_thread_quitting,
            start_mission,
            )
        self.start_mission_thread.start()
```

src/isar/robot/robot.py
```diff
@@ -0,0 +1,90 @@
...
+                break
+            start_mission_or_task = check_for_event(
+                self.queues.state_machine_start_mission
+            )
+            if start_mission_or_task is not None:
+                if (
+                    self.start_mission_thread is not None
+                    and self.start_mission_thread.is_alive()
+                ):
+                    self.start_mission_thread.join()
```
oysand, line 67: Does this allow a new mission to start as long as the previously started mission regardless of it is still running? Should this throw an error/report erro instead?

src/isar/robot/robot.py
andchiind: Nice. I'll do this. A good step towards this: https://github.com/equinor/isar/issues/715

andchiind: Although I might make the parameter the specific queue we are checking so we don't have to jump around so much to find that info. We don't need the rest of the queue anyways.

src/isar/robot/robot.py
andchiind: I think I will implement @tsundvoll suggestion below first and then we can reevaluate this.

Christdej (changes requested):

src/isar/robot/robot_status.py
```diff
@@ -0,0 +1,45 @@
...
+from isar.models.communication.queues.queues import Queues
+from isar.robot.robot_queue_utils import trigger_event
+from isar.services.utilities.threaded_request import (
+    ThreadedRequest,
+)
+from robot_interface.models.mission.status import RobotStatus
+from robot_interface.robot_interface import RobotInterface
+
+
+class RobotStatusThread(Thread):
```
oysand, line 15: Should this be running at all times or be requested updates on for the state machine?

src/isar/state_machine/transitions/stop.py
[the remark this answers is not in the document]
andchiind: That is from how it was before, but I have an issue on improving it. https://github.com/equinor/isar/issues/712
Nevertheless though, we need somewhere to check for the success of stopping a mission, so having a stop state makes sense, but we also need to have a transition to trigger the stop.

tests/isar/state_machine/test_state_machine.py
```diff
@@ -65,6 +85,13 @@ def uploader_thread(injector) -> UploaderThread:
     return UploaderThread(injector=injector)
 
 
+@pytest.fixture
+def robot_service_thread(injector):
+    robot_service_thread: RobotServiceThread = RobotServiceThread(injector=injector)
+    yield robot_service_thread
+    robot_service_thread.teardown()
+
+
```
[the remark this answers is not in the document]
andchiind: We can do that as a larger refactor, as this is currently not done for the other fixtures.

.github/workflows/pythonpackage.yml
andchiind: Because when I ran mypy it complained that it was not checking some things that could be relevant, and I though that it then makes sense to run mypy in a more strict way. It only affected the new code, but I think it makes sense to include this going forward.

src/isar/models/communication/queues/queues.py
andchiind: No, once the state machine consumes the event the API thread can no longer track it, but that is not some behaviour which has changed in the PR.

tests/isar/state_machine/test_state_machine.py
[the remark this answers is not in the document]
Christdej: I think we should still put it in conftest.py, and future fixtures, that we minimize the refactor we have to do once we want to move all fixtures to conftest.py

src/isar/models/communication/queues/queues.py
Christdej: I agree, more of a curiosity from my side.

src/isar/state_machine/transitions/stop.py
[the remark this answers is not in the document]
Christdej: Good that it is documented in a issue.

src/isar/robot/robot.py
andchiind: I could delete the placeholder though and refer to it in the issue only

.github/workflows/pythonpackage.yml
andchiind: Here is how it looks like without it
image

src/isar/robot/robot_status.py
andchiind: Yes, I can update it

andchiind: Seems like the BlockedProtectiveStop re-added the robot status thread in Idle in a rebase. I will remove it again and only use this service.

mrica-equinor (approved):
LGTM :) 
Further refactors and inprovements can be taken in another PR since this is also blocking Øystein and other people from using this new version of teh code.

tests/isar/state_machine/test_state_machine.py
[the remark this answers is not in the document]
andchiind: https://github.com/equinor/isar/issues/731

oysand:
The state diagram should be updated as well

src/isar/robot/robot.py
```diff
@@ -0,0 +1,89 @@
...
+from isar.robot.robot_start_mission import RobotStartMissionThread
+from isar.robot.robot_status import RobotStatusThread
+from isar.robot.robot_task_status import RobotTaskStatusThread
+from robot_interface.robot_interface import RobotInterface
+
+
+class Robot(object):
+
+    @inject
+    def __init__(self, queues: Queues, robot: RobotInterface):
```
tsundvoll, line 17:
```suggestion
    def __init__(self, queues: Queues, robot: RobotInterface) -> None:
```

src/isar/robot/robot_queue_utils.py
```diff
@@ -0,0 +1,26 @@
+import queue
+from typing import Any
+from isar.models.communication.queues.status_queue import StatusQueue
+from isar.models.communication.queues.queue_io import QueueIO
+
+
+def trigger_event(queueio: QueueIO, data: Any = None) -> Any:
```
tsundvoll, line 7:
```suggestion
def trigger_event(queueio: QueueIO, data: Any = None) -> None:
```

src/isar/robot/robot_queue_utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def check_shared_state(queueio: StatusQueue):
+    try:
+        return queueio.check()
+    except queue.Empty:
+        return None
+
+
+def update_shared_state(queueio: StatusQueue, data: Any = None) -> Any:
```
tsundvoll, line 18:
```suggestion
def update_shared_state(queueio: StatusQueue, data: Any = None) -> None:
```

src/isar/robot/robot_start_mission.py
```diff
@@ -0,0 +1,77 @@
...
+from robot_interface.models.exceptions.robot_exceptions import (
+    ErrorMessage,
+    RobotException,
+)
+from robot_interface.robot_interface import RobotInterface
+
+from robot_interface.models.mission.mission import Mission
+from robot_interface.models.exceptions.robot_exceptions import (
+    RobotInfeasibleMissionException,
+)
```
tsundvoll, line 19: Sort the imports

andchiind: I ran it and it changed many more files than my PR already changes, so I would rather run it in another PR. There is already enough issues with merge conflicts in this PR as-is.

tsundvoll (approved):
Good work, looking forward to see it in dev!

PR #694, opened 2025-01-15
Add blocked protective stop state

oysand:
- [x] A test has been written

### DIFF
--- a/src/isar/state_machine/state_machine.py
+++ b/src/isar/state_machine/state_machine.py
@@ -24,6 +24,7 @@ from isar.state_machine.states.initiate import Initiate
 from isar.state_machine.states.monitor import Monitor
 from isar.state_machine.states.off import Off
 from isar.state_machine.states.offline import Offline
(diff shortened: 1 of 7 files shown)


src/isar/state_machine/state_machine.py
```diff
@@ -227,6 +230,18 @@ def __init__(
                     "dest": self.idle_state,
                     "before": self._online,
                 },
+                {
+                    "trigger": "robot_protective_stop_engaged",
+                    "source": [self.idle_state],
```
Christdej, line 235: What will happend if protective stop is engaged while state machine is in Monitor or other states?

oysand: It depends on the situation. If the robot is in a mission (monitor) it will continue to monitor the mission until the mission fails or is cancelled. Then once the robot mission fail it will go to idle and transition to BlockedProtectiveStop. This way if someone activate protective stop for a bit and release it again the mission can continue without isar needing to do any action

Christdej (approved):
LGTM

PR #699, opened 2025-01-24
Remove task wise mode

oysand:
- [ ] A test has been written


src/isar/state_machine/states/monitor.py
```diff
@@ -133,28 +130,22 @@ def _run(self) -> None:
...
+                    task=self.state_machine.current_task
+                )
 
-                    self.state_machine.iterate_current_task()
-                    if self.state_machine.current_task is None:
-                        transition = self.state_machine.full_mission_finished  # type: ignore
-                        break
+                self.state_machine.iterate_current_task()
+                if self.state_machine.current_task is None:
+                    transition = self.state_machine.full_mission_finished  # type: ignore
```
oysand, line 141: rename to mission_finished

Eddasol:
There are a lot of taskwise related comments in robot_interface that can be removed as well. Like this one:
```
This function should be used in combination with the mission_status function
if the robot is designed to run full missions and not in a taskwise
configuration.
```

src/robot_interface/robot_interface.py
```diff
@@ -71,7 +61,7 @@ def initiate_task(self, task: Task) -> None:
             of the task until the number of allowed retries is exceeded before the task
             will be marked as failed and the mission cancelled
         NotImplementedError
```
Christdej, line 63: A few lines above this function still says "RobotInfeasibleTaskException" which I believe is no longer possible? Also you can delete this exception from robot_exception.py

Christdej (changes requested):

src/robot_interface/robot_interface.py
oysand: I think this is still a reasonable exception as we want to keep initiate_task for scheduling retrun to home from isar directly

andchiind:
It might be easier to merge this PR first: https://github.com/equinor/isar/pull/689 since it isolates all the task-wise code into specific transitions. Instead of deleting specific lines in various states, all we would need to do is delete the relevant transitions (and remove the Task option in starting missions).

src/robot_interface/robot_interface.py
oysand: I would argue so, yes

Christdej (approved):
LGTM

oysand:
> It might be easier to merge this PR first: #689 since it isolates all the task-wise code into specific transitions. Instead of deleting specific lines in various states, all we would need to do is delete the relevant transitions (and remove the Task option in starting missions).

As this change entails less change in isar, we will take this in first and continue with it afterwards

PR #725, opened 2025-02-19
Add return home

oysand:
- [x] A test has been written


oysand:
To be combined with: https://github.com/equinor/isar-robot/pull/206

oysand:
To be combined with: https://github.com/equinor/flotilla/pull/2041

src/isar/state_machine/states/await_next_mission.py
```diff
@@ -0,0 +1,56 @@
...
+from transitions import State
+
+from isar.config.settings import settings
+from isar.models.communication.message import StartMissionMessage
+
+if TYPE_CHECKING:
+    from isar.state_machine.state_machine import StateMachine
+
+
+class AwaitNextMission(State):
```
andchiind, line 14: Why do we need both this state and Idle?

src/isar/state_machine/states/docked.py
```diff
@@ -0,0 +1,91 @@
...
+    ThreadedRequestNotFinishedError,
+)
+from robot_interface.models.exceptions.robot_exceptions import RobotException
+from robot_interface.models.mission.status import RobotStatus
+
+if TYPE_CHECKING:
+    from isar.state_machine.state_machine import StateMachine
+
+
+class Docked(State):
```
andchiind, line 20: If we have a docked state that implies that the robot is not docked in Idle. Should we rename Idle to reflect this?

oysand: Renamed Docked to IdleDocked to indicate that it is similar to idle

src/isar/state_machine/states/await_next_mission.py
oysand: The await next mission state is the state that implements the triggering of the auto return home once all the missions are finished

oysand: Ref discussions on state diagram

src/isar/state_machine/states/docked.py
oysand: Changed to home

Christdej (changes requested):
Some changes after reviewing approximately half of the PR

src/isar/state_machine/state_machine.py
```diff
@@ -98,114 +86,54 @@ def __init__(
...
-                },
-            ]
+        self.machine = Machine(
+            self, states=self.states, initial="unknown_status", queued=True
         )
 
+        self.transitions: List[dict] = []
+
+        for transition in get_mission_transitions(self):
+            self.transitions.append(transition)
```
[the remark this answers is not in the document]
oysand: Fixed

src/isar/robot/robot.py
```diff
@@ -95,6 +96,14 @@ def _check_and_handle_stop_mission(self, event: Queue) -> None:
                 self.stop_mission_thread_thread is not None
                 and self.start_mission_thread.is_alive()
             ):
+                error_description = "Received stop mission event mission is trying to start. Aborting stop attempt."
```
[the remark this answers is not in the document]
oysand: Fixed

README.md
```diff
@@ -195,16 +195,15 @@ The state machine is based on the [transitions](https://github.com/pytransitions
...
 States.Monitor,
 States.Paused,
 ```
 
-indicates that the state machine is already running. For running a mission the state machine need to be in the state
+indicates that the state machine is already running. For running a mission the state machine need to be in the states
 
 ```
-States.Idle
+States.Home, States.RobotStandingStill, States.AwaitNextMission or States.ReturningHome
```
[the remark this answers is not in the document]
oysand: Yeah, I don't have a strong preference between Idle and RobotStandingStill

docs/extended_state_machine_diagram.png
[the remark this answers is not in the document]
oysand: Added all diagrams to the readme. I'm not quite sure if we need both the core mission behavior and the extended functionality that includes the return home?

src/isar/apis/api.py
```diff
@@ -155,6 +155,28 @@ def _create_scheduler_router(self) -> APIRouter:
...
+            endpoint=self.scheduling_controller.return_home,
+            methods=["POST"],
+            dependencies=[authentication_dependency],
+            summary="Start the mission provided in JSON format",
+            responses={
+                HTTPStatus.OK.value: {
+                    "description": "Mission succesfully started",
+                    "model": StartMissionResponse,
+                },
+                HTTPStatus.UNPROCESSABLE_ENTITY.value: {
```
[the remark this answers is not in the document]
oysand: Do you mean that the robot cannot run a return home mission or that it is currently unable to run a return home mission? I think we need to enforce that the robot package need to handle return home mission, though it can be handled differently when it comes to whether or not it goes to a specific position or actually return home

src/isar/apis/schedule/scheduling_controller.py
```diff
@@ -98,6 +103,23 @@ def start_mission(
...
+        )
+
+        mission = Mission(
+            tasks=[ReturnToHome()],
+            name="Return Home",
+        )
+
+        self.logger.info(f"Starting mission: {mission.id}")
+        self.scheduling_utilities.start_mission(mission=mission)
+        return self._api_response(mission)
```
andchiind, line 121: I see that you are sending return home as a normal mission and then checking it in the state machine what kind of mission you get and reacting accordingly. Wouldn't it be cleaner to instead have this be a different event sent to the state machine which specifically tells it to start a return home mission? That way we can have a specific event handler for the event in robotStandingStill for example, instead of reusing other events.

src/isar/state_machine/states/await_next_mission.py
```diff
@@ -0,0 +1,76 @@
...
+            return True
+        return False
+
+    def _check_and_handle_start_mission_event(
+        self, event: Queue[StartMissionMessage]
+    ) -> bool:
+        start_mission: Optional[StartMissionMessage] = check_for_event(event)
+        if start_mission:
+            self.state_machine.start_mission(mission=start_mission.mission)
+            if self.state_machine.current_mission._is_return_to_home_mission():
```
andchiind, line 48: Related to my previous comment

src/isar/state_machine/states/home.py
```diff
@@ -0,0 +1,71 @@
...
+    def stop(self) -> None:
+        return
+
+    def _check_and_handle_start_mission_event(
+        self, event: Queue[StartMissionMessage]
+    ) -> bool:
+        start_mission: Optional[StartMissionMessage] = check_for_event(event)
+        if start_mission:
+            self.state_machine.start_mission(mission=start_mission.mission)
+            if self.state_machine.current_mission._is_return_to_home_mission():
```
andchiind, line 41: Related to my previous comment

src/isar/state_machine/states/robot_standing_still.py
```diff
@@ -45,20 +46,22 @@ def _check_and_handle_start_mission_event(
         start_mission: Optional[StartMissionMessage] = check_for_event(event)
         if start_mission:
             self.state_machine.start_mission(mission=start_mission.mission)
-            self.state_machine.request_mission_start()  # type: ignore
-            return True
+            if self.state_machine.current_mission._is_return_to_home_mission():
```
andchiind, line 49: Related to my previous comment

andchiind (changes requested):

src/isar/apis/schedule/scheduling_controller.py
oysand: Agreed, introduced a return_home event that address this. The checking of whether it is a return home mission is kept as it is currently used so that stopping can differentiate what kind of mission have been stopped

src/isar/robot/robot.py
```diff
@@ -95,6 +96,14 @@ def _check_and_handle_stop_mission(self, event: Queue) -> None:
                 self.stop_mission_thread_thread is not None
                 and self.start_mission_thread.is_alive()
             ):
+                error_description = "Received stop mission event mission while trying to start a mission. Aborting stop attempt."
```
mrica-equinor, line 99: error_description = "Received stop mission event while trying to start a mission. Aborting stop attempt."

src/isar/services/utilities/scheduling_utilities.py
```diff
@@ -119,19 +119,47 @@ def verify_robot_capable_of_mission(
...
+
+        Raises
+        ------
+        HTTPException 409 Conflict
+            If state machine is not home, robot standing still or awaiting next mission
+            and therefore cannot start a new return home mission
+        """
+        if (
+            state == States.RobotStandingStill
+            or state == States.Home
```
mrica-equinor, line 155: If the robot is home does it make sense to schedule a return to home mission? Is this in case you want to move it between docking stations?

src/isar/state_machine/states/home.py
```diff
@@ -0,0 +1,79 @@
...
+        start_mission: Optional[StartMissionMessage] = check_for_event(event)
+        if start_mission:
+            self.state_machine.start_mission(mission=start_mission.mission)
+            self.state_machine.request_mission_start()  # type: ignore
+            return True
+        return False
+
+    def _check_and_handle_return_home_event(self, event: Queue) -> bool:
+        if check_for_event(event):
+            event.output.put(True)
```
mrica-equinor, line 47: return_home: Optional[ReturnHomeMessage] = check_for_event(event)
if return_home:
 ....
 
 suggest this for consistency with other bits of the code

src/isar/state_machine/states/await_next_mission.py
mrica-equinor: Why was this better than having just idle and say if it doesn't receive a mission within a few minutes, then return home? is it to account for future functionality in which the robot might actually need to be standing in front of something for live stream or another use case?

tests/test_data/test_mission_working.json
```diff
@@ -2,25 +2,6 @@
   "id": "1",
   "name": "Well defined mission",
```
mrica-equinor, line 3: Well defined return to home mission

oysand: Removed the return home reference

src/isar/state_machine/states/home.py
oysand: There is no message on this event since the mission is created in the transition. This is more in line with other such event, for instance stop

src/isar/services/utilities/scheduling_utilities.py
oysand: Ah, that is not necessary

oysand: Though if it is in the home state by mistake and we would like to schedule the mission again it could be relevant to have the option of sending it even though it is already home

src/isar/state_machine/transitions/robot_status.py
```diff
@@ -0,0 +1,73 @@
...
+    is_offline,
+)
+from isar.state_machine.transitions.functions.utils import def_transition
+
+if TYPE_CHECKING:
+    from isar.state_machine.state_machine import StateMachine
+
+
+def get_robot_status_transitions(state_machine: "StateMachine") -> List[dict]:
+    robot_status_transitions: List[dict] = [
```
andchiind, line 16: This does kind of seem like an abuse of the conditional transitions functionality, but if it works then I guess it's fine.

andchiind: It does however seem to take away from the overall functionality of having a state machine, since here it is not state machine describing the state of the system, but the robot interface implementation doing so. It seems like the logic is a bit reversed in my opinion to make this the default method of updating the state, and I would argue it should be the state machine impacting the robot status and not the other way around. Again, I will not let this stop the PR though.

src/isar/services/utilities/scheduling_utilities.py
oysand: It would not make so that flotilla would think that ISAR is running a return home mission when it isn't. With the changes in https://github.com/equinor/flotilla/pull/2041, Flotilla will not keep track of the return home mission, only base it on the robot status reported by ISAR. This is to account for cases where the robot package implementation have missed an edge case and thinks it is home when it is not. That way Flotilla can tell ISAR to return home even if ISAR thinks it is already home.

src/isar/state_machine/transitions/functions/return_home.py
```diff
@@ -0,0 +1,38 @@
...
+
+def start_return_home_mission(state_machine: "StateMachine") -> bool:
+    state_machine.start_mission(
+        Mission(
+            tasks=[ReturnToHome()],
+            name="Return Home",
+        )
+    )
+    return True
+
```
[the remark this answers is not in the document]
oysand: We could, but then we would need to have the robot service distinguish between return home missions and regular missions. This way the robot service does not need to know the difference

src/isar/state_machine/transitions/mission.py
```diff
@@ -0,0 +1,181 @@
...
+        {
+            "trigger": "mission_stopped",
+            "source": state_machine.stopping_state,
+            "dest": state_machine.await_next_mission_state,
+            "before": def_transition(state_machine, stop_mission_cleanup),
+        },
+        {
+            "trigger": "return_home_mission_stopped",
+            "source": state_machine.stopping_state,
+            "dest": state_machine.offline_state,
```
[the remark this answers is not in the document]
oysand: It is to avoid using the unkown_status as a hub state and avoid having to have stopping determine which transitions it should utilize to arrive in the most accurate waiting state. Since the robot can turn offline while stopping, the transition account for this and the unkown status will be as a catch all for the robot statuses which are not linked to a dedicated waiting state

src/isar/state_machine/states/monitor.py
```diff
@@ -55,10 +55,15 @@ def _check_and_handle_pause_mission_event(self, event: Queue) -> bool:
...
-                self.state_machine.mission_ongoing = True
+    def _check_and_handle_mission_started_event(self, event: Queue) -> bool:
+        if self.state_machine.mission_ongoing:
+            return False
+
+        if check_for_event(event):
+            self.state_machine.mission_ongoing = True
+            return False
+
+        return True
```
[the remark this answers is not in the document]
oysand: It does differ a bit in the sense that this check will skip the rest if it has not yet occured. We could change it, but then it would be an if not check in the while loop: 
image

README.md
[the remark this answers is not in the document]
oysand: I think RobotStandingStill can be less useful over time as the state machine better represents all the cases that the robot system can be. At some point it can probably be removed, however for now we are quite a ways off being able to remove it

src/isar/state_machine/states/await_next_mission.py
oysand: It is to differentiate between return home as part of regular mission behaviour. The current implementation is that we return home after 10 seconds (as default) if we do not receive a new mission. If we where to combine RobotStandingStill with AwaitNextMission with the current implementation of ISAR it would result in the restarting of the return home after 10 seconds if the return home is for any reason stopped

src/isar/state_machine/transitions/robot_status.py
oysand: Agreed, let's address in later refactor

src/isar/state_machine/transitions/mission.py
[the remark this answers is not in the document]
oysand: Sending return home stopped only to robot standing still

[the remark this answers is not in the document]
oysand: Addressed

src/isar/state_machine/transitions/robot_status.py
```diff
@@ -0,0 +1,73 @@
...
+        {
+            "trigger": "robot_status_changed",
+            "source": [
+                state_machine.robot_standing_still_state,
+                state_machine.home_state,
+                state_machine.blocked_protective_stopping_state,
+                state_machine.offline_state,
+                state_machine.unknown_status_state,
+            ],
+            "dest": state_machine.unknown_status_state,
```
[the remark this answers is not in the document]
oysand: https://github.com/equinor/isar/issues/762

andchiind (approved):
I think my main concerns have been addressed now, so I will approve it, but I would suggest you don't merge the PR until the other blocking reviews are changed to approved as well. We have discussed some limitations with the PR, but these will be addressed in future PRs, as this PR has been tested and works as-is.

Christdej (approved):
LGTM

PR #757, opened 2025-03-27
Use dependency-injector instead of injector

Christdej:
- [ ] A test has been written


pyproject.toml
```diff
@@ -34,7 +34,7 @@ dependencies = [
     "dacite",
     "fastapi-azure-auth",
     "fastapi",
-    "injector",
+    "dependency_injector",
```
tsundvoll, line 37:
```suggestion
    "dependency-injector",
```

tsundvoll:
The tests pass on my machine now.

I do have a problem when running isar-start, though. It seems like the SLIMM storage handler is triggered, even though I have disabled it in settings. Needs to be investigated before merging

tsundvoll:
This PR also disables daemon threads (it does not create threads as daemon threads) in tests.

Investigate if we can do the same in scripts.py when running for real. (Remove daemon=True)

src/isar/robot/robot.py
```diff
@@ -107,10 +106,7 @@ def run(self) -> None:
         )
         self.robot_status_thread.start()
 
-        while True:
-            if self.signal_thread_quitting.is_set():
-                break
-
+        while not self.signal_thread_quitting.wait(0):
```
tsundvoll, line 109: This trick makes the tests more responsible when they fail (instead of hanging, they terminate faster)

src/isar/robot/robot_status.py
```diff
@@ -21,7 +20,7 @@ def __init__(
         self.robot: RobotInterface = robot
         self.signal_thread_quitting: Event = signal_thread_quitting
         self.last_robot_status_poll_time: float = time.time()
-        Thread.__init__(self, name="Robot status thread", daemon=True)
+        Thread.__init__(self, name="Robot status thread")
```
tsundvoll, line 23: We can also consider not running daemon threads everywhere

src/isar/robot/robot_status.py
```diff
@@ -35,10 +34,10 @@ def _is_ready_to_poll_for_status(self) -> bool:
         )
 
     def run(self):
-        while True:
-            if self.signal_thread_quitting.is_set():
-                break
+        if self.signal_thread_quitting.is_set():
+            return
 
+        while not self.signal_thread_quitting.wait(0.001):
```
tsundvoll, line 40: It was necessary to check both before the while loop and in the condition for one of the tests. I think it was because the quitting signal was sent even before the run function was entered

PR #840, opened 2025-07-09
Use received mission id, if any

Eddasol:
- [ ] A test has been written

### DIFF
--- a/src/isar/apis/models/start_mission_definition.py
+++ b/src/isar/apis/models/start_mission_definition.py
@@ -53,6 +53,7 @@ class StartMissionTaskDefinition(BaseModel):
 
 
 class StartMissionDefinition(BaseModel):
+    id: Optional[str] = None
     tasks: List[StartMissionTaskDefinition]
     name: Optional[str] = None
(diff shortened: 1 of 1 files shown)


src/isar/apis/models/start_mission_definition.py
```diff
@@ -80,7 +81,10 @@ def to_isar_mission(
     if start_mission_definition.start_pose:
         start_pose = start_mission_definition.start_pose.to_alitra_pose()
 
+    id = start_mission_definition.id if start_mission_definition.id else uuid4_string()
+
     return Mission(
+        id=id,
```
tsundvoll, line 87: I agree to this change, and the line 84 here can be removed, as the default value will be set to a uuid4_string in the constructor of the StartMissionDefinition class.

We might want to do input validation on that id to verify that it is an uuid coming in.

```suggestion
    return Mission(
        id=start_mission_definition.id,
```

Eddasol: If we want validation we should make a separate issue for it as that would make sense to implement other places as well, like task id

PR #846, opened 2025-07-11
Allow robot to disregard return to home when home

Eddasol:
- [ ] A test has been written

### DIFF
--- a/src/isar/models/events.py
+++ b/src/isar/models/events.py
@@ -165,6 +165,7 @@ class RobotServiceEvents:
         self.request_inspection_upload: Event[Tuple[TASKS, Mission]] = Event(
             "request_inspection_upload"
         )
+        self.robot_already_home: Event[bool] = Event("robot_already_home")
 
 
 class SharedState:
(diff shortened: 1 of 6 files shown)


andchiind (commented):
I might be a little confused, but could we not just return a 408 Conflict if we receive a request to return home while we are in the Home state?

Eddasol:
> I might be a little confused, but could we not just return a 408 Conflict if we receive a request to return home while we are in the Home state?

The suggestion was that this should be handled in the robot package as some robots know if the robot is home and see no need for retry, while other robots assume home and can benefit from rescheduling a return to home

src/isar/models/events.py
```diff
@@ -126,6 +126,9 @@ def __init__(self) -> None:
         self.release_from_maintenance_mode: APIEvent[bool, bool] = APIEvent(
             "release_from_maintenance_mode"
         )
+        self.already_returned_home: APIEvent[bool, bool] = APIEvent(
+            "already_returned_home"
+        )
```
andchiind:
```suggestion
```

PR #863, opened 2025-07-30
Remove queue utilities file

andchiind:
- [ ] A test has been written


src/isar/models/communication/queues/queue_timeout_error.py
```diff
@@ -1,2 +1,2 @@
-class QueueTimeoutError(Exception):
+class EventTimeoutError(Exception):
```
tsundvoll, line 1: Class name and filename should probably match. Maybe the exception might as well be placed inside the events.py file?

tsundvoll: Actually, the message.py can also be removed, and the `StartMissionMessage` can be removed and we can use the `Mission` class directly instead.

That means the /communication folder can also go and the events.py can live just under the /models folder

andchiind: Done 👍

tsundvoll (approved):
LGTM 👍

PR #875, opened 2025-08-06
Rename variables for increased readability

### LINKED ISSUE #874: Refactors

- [x] Rename put_start_mission_on_queue to acknowledge_mission
  - More precise naming
- [x] Rename initiate_mission in start_mission.py to prepare_state_machine_before_mission or similar
  - Does not really initiate the mission, but sets some variables in the state machine and publishes message 
  - Also initiate_mission is already used in robot_interface.py, actually initiating mission
- [ ] We can consider a Preparing Mission State in addition to the Running Mission State (currently Monitor)
  - This can simplify some interactions
    - We will not accept other events while preparing
    - Easier to cancel the preparation before the mission has started (or attempted to be started)
- [ ] Convert the global variable state_machine.mission_ongoing into an actual state
  - Related to the point above. If we have a dedicated state where the mission is actually running, we don't need the extra global variable
- [x] Rename in API Events
  - Input -> Request
  - Output -> Response (or better: Acknowledge)   

This is mostly renaming and should not change functionality

More readable code

akustn:
- [ ] A test has been written

Partially fixes #874

### DIFF
--- a/src/isar/state_machine/transitions/functions/stop.py
+++ b/src/isar/state_machine/transitions/functions/stop.py
@@ -31,7 +31,9 @@ def stop_mission_cleanup(state_machine: "StateMachine") -> bool:
     stopped_mission_response: ControlMissionResponse = (
         state_machine._make_control_mission_response()
     )
-    state_machine.events.api_requests.stop_mission.output.put(stopped_mission_response)
+    state_machine.events.api_requests.stop_mission.response.put(
+        stopped_mission_response
+    )
     state_machine.publish_task_status(task=state_machine.current_task)
     state_machine._finalize()
     return True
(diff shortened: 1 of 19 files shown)


src/isar/state_machine/transitions/functions/stop.py
```diff
@@ -31,7 +31,9 @@ def stop_mission_cleanup(state_machine: "StateMachine") -> bool:
     stopped_mission_response: ControlMissionResponse = (
         state_machine._make_control_mission_response()
     )
-    state_machine.events.api_requests.stop_mission.output.put(stopped_mission_response)
+    state_machine.events.api_requests.stop_mission.response.put(
+        stopped_mission_response
+    )
```
tsundvoll, line 36: Good example why this renaming is a good idea; the input variable was already named response :)

Eddasol:
Suggested improvement: Rename PR to something more descriptive for better release notes. Something like: Rename variables for increased readability

PR #898, opened 2025-08-28
Make robot charge before running missions when the battery is low

andchiind:
The reason this PR makes it so that queue timeouts count as a conflict in HTTP requests is that such timeouts are possible when an API request is received at the exact same time that the state machine moves to a state that does not have an event handler for that API request. An alternative solution to this timeout would be for all states to accept all API requests and then return 409 CONFLICT replies to non-relevant ones, but this solution effectively achieves the same result.

Here is an instance of such a timing problem occuring:
image
This issue is not introduced in this PR, I just happened to notice it while debugging it.

- [x] A test has been written


Christdej (approved):
LGTM

PR #927, opened 2025-09-26
Make task status into telemetry

andchiind:
Basically, this PR makes task status no longer relevant to the state machine. It is simply reported to MQTT in the robot service, and then instead we rely solely on mission_status for the actual state machine event handlers.

- [x] A test has been written

### DIFF
--- a/src/isar/robot/robot.py
+++ b/src/isar/robot/robot.py
@@ -47,6 +47,7 @@ class Robot(object):
         self.pause_mission_thread: Optional[RobotPauseMissionThread] = None
         self.upload_inspection_threads: List[RobotUploadInspectionThread] = []
         self.signal_thread_quitting: ThreadEvent = ThreadEvent()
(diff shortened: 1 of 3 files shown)


andchiind:
Note that I am deleting the "states" folder in the testing project. This is only because all the tests defined there became irrelevant. I do think splitting up the eventhandler tests by states is a good idea, so we should re-add this folder when we do so.

src/isar/robot/robot.py
```diff
@@ -149,6 +167,28 @@ def _pause_mission_request_handler(self, event: Event[bool]) -> None:
             )
             self.pause_mission_thread.start()
 
+    def _upload_inspection_event_handler(
+        self, event: Event[Tuple[TASKS, Mission]]
+    ) -> None:
+        upload_request = event.consume_event()
```
oysand: Consider increasing the queue size

PR #964, opened 2025-10-30
Make transition code more readable/maintainable

andchiind:
The transition library is great for a lot of things, but not readability. The aim of this PR is to move transition code to places where it's used, and to delete unnecessary transition code. This means that we can easily see what code will be ran in a state's transition, simply by looking at the state code. Previously to find what code would be ran you would need to search for the string name of the transition (which probably wouldn't be unique), then find all the references to the functions being ran in the transition definition, and then look up those function definitions in turn.

Even though there might be concerns about this resulting in duplicate code, here are my two cents:
1. this PR deletes more code than it adds. Using the transition function setup was not more succint.
2. duplicate code is not always bad. This is especially true in this case where changing a transition function could easily result in unforeseen changes elsewhere, since there was no simple way of telling where the function was being used, and within what context. Just because some code seems duplicate now does not mean that that code block will be duplicate forever.

Once we merge this PR: https://github.com/equinor/isar/pull/966, we will be able to delete all transition functions and related files, significantly improving code readability. At that point, all information related to state transitions will be stored in the states themselves.

I am open to criticism if there are objections to this PR. I understand that it feels clean to group together code for transitions that are used by several states. If people are unhappy with having to repeat some code when adding a new transition, I suggest using the on_entry and on_transition functions that I have added instead. These act very similar to the transition functions, but they are call each time we enter the state, so they are more general, and additionally they are defined in the states themselves, along with whatever temporary state they might manipulate, making them much less obfuscated. Currently they are only used for the return home retry counter, which is itself a temporary fix, and we might therefore remove this functionality in the future.

In an ideal world, the transition functions themselves would be normal python functions. This way we could accept parameters that are required when transitioning, and then we would easily be able to find the code that is executed in the transition by looking at the transition function itself. So when starting a mission we could pass the mission itself in the transition function, and when stopping a mission we could pass the mission ID to the stopping state, so that it could maintain this information without needing any shared state for it. Maybe in the future we can extend the library to achieve this?

- [x] A test has been written


oysand (changes requested):
Do not expand await next mission

src/isar/state_machine/states/going_to_lockdown.py
```diff
@@ -54,6 +54,7 @@ def _mission_status_event_handler(
                 if mission_status != MissionStatus.Successful:
                     return state_machine.lockdown_mission_failed  # type: ignore
 
+                state_machine.print_transitions()
```
andchiind: Tbf I'm not sure if we need to print these out anymore. They don't really provide any new information. As far as I can tell, the transition list is only used for the older integration tests.

src/isar/eventhandlers/eventhandler.py
```diff
@@ -38,6 +38,8 @@ def __init__(
         state_name: str,
         event_handler_mappings: List[EventHandlerMapping],
         timers: List[TimeoutHandlerMapping] = [],
+        on_entry: Optional[Callable[[], None]] = None,
+        on_transition: Optional[Callable[[], None]] = None,
```
andchiind: These might be removed/changed in a future refactor, once we have extended/replaced the transitions library. They are useful now to ensure that certain conditions have been met when entering a state, but this can be replaced by more systematic testing.

nicholasdalhaug (approved):
LGTM

PR #1003, opened 2025-11-25
Respond quickly when stopping return home

andchiind:
In this PR we don't wait for the return home to be stopped before responding to the start mission request. This avoids the request timing out due to the stopping of the return home taking a long time. This can happen when we first need to wait for the return home mission to start before stopping it.

Flotilla should anyways not assume the mission is ongoing until we have confirmation of this on MQTT.

This PR also fixes an occasional bug where the mission status was set in robot_monitor_thread.py right as we were stopping the mission, and where this value was then not cleared before moving on. We now clear this event when we get a confirmation that stopping has succeeded. The mission status will anyways be reported on MQTT, but we don't use it in the stopping state to confirm that the mission has stopped, we only rely on the 'mission_successfully_stopped' event.

- [ ] A test has been written

### DIFF
--- a/src/isar/robot/robot.py
+++ b/src/isar/robot/robot.py
@@ -140,6 +140,12 @@ class Robot(object):
             self.stop_mission_thread is not None
             and not self.stop_mission_thread.is_alive()
         ):
+            if self.monitor_mission_thread is not None:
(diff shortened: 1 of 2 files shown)


oysand (approved):
LGTM

PR #1006, opened 2025-11-26
Moving state machine test into separate files

### LINKED ISSUE #928: Move state_machine tests into relevant state files
Most new tests in the statemachine are unit tests which test the functionality of specific eventhandlers (which is a good thing). I suggest grouping these tests into files based on which state these handlers belong to. So a test that checks functionality in eventhandler "send_to_lockdown_event" in returning home state (for example), should be moved to a test_returning_home_state.py test file.

It would make the test files easier to navigate, and would make it more obvious where we are lacking tests.

N/A

akustn:
Closes #928 

- [ ] A test has been written


Christdej (approved):
LGTM

PR #1034, opened 2025-12-16
Report mission NotStarted

### LINKED ISSUE #1031: Report mission status when the mission is trying to start
Currently we only report mission status once the mission has started. To get better feedback on MQTT we should send an event when we are trying to start the mission. This can either be to reuse the mission status event to just send status "NotStarted" or we can make a new event such as "initialising mission" or something like that. This will let any listeners know what mission is being worked on before it is inProgress.

It will give faster response

N/A

andchiind:
Closes #1031 

See https://github.com/equinor/flotilla/pull/2515

- [ ] A test has been written

### DIFF
--- a/src/isar/robot/robot.py
+++ b/src/isar/robot/robot.py
@@ -193,6 +193,8 @@ class Robot(object):
                 )
                 self.start_mission_thread.join()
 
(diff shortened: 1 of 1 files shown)


Eddasol (approved):
LGTM :)

PR #1038, opened 2025-12-17
Handle that robot can auto pause a mission

Eddasol:
NB: This only handles auto-pausing if it occurs during monitor or returning home. If auto-pausing occur during goingToRecharge or goingToLockdown, it wont be possible to resume as these states dont have a corresponding version of paused. Could be implemented, but as this is a workaround that can most likely be removed after some months, it might not be worth the effort.

- [ ] A test has been written

### DIFF
--- a/src/isar/state_machine/states/monitor.py
+++ b/src/isar/state_machine/states/monitor.py
@@ -9,7 +9,7 @@ from isar.state_machine.utils.common_event_handlers import (
     stop_mission_event_handler,
 )
 from robot_interface.models.exceptions.robot_exceptions import ErrorMessage
-from robot_interface.models.mission.status import MissionStatus
+from robot_interface.models.mission.status import MissionStatus, RobotStatus
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


src/isar/state_machine/states/monitor.py
```diff
@@ -102,6 +102,22 @@ def _mission_failed_event_handler(
...
 
+        def _robot_status_event_handler(
+            status_changed_event: Event[bool],
+        ) -> Optional[Callable]:
+            has_changed = status_changed_event.consume_event()
+            if not has_changed:
+                return None
+            robot_status: Optional[RobotStatus] = shared_state.robot_status.check()
+            if robot_status == RobotStatus.Busy:
+                return None
```
andchiind:
```suggestion
```

andchiind (commented):
Would it be possible to detect pausing using the mission status instead of the robot status? It would be nice to keep robot status separate from mission execution.

Eddasol:
Will be fixed directly in robot package instead

PR #1049, opened 2025-12-29
Remove transitions library and simplify event handlers

andchiind:
This PR contains a working solution that does not use the transitions library and instead handles transitions as normal python functions, including precise type hinting. This avoids having a disconnect between transitions and states, where transitions are referenced as strings which need to be looked up to validate. Each state now directly returns the instance of the next state when it transitions. States themselves effectively become immutable objects, as we create new ones instead of instantiating them all once when starting the state machine. This makes the code more functional and testable, avoiding lingering side effects.

image

Example where mypy complains about an incorrect transition:
image

image

The new system also allows transition functions to take arguments (with type hinting). This means that if a certain value is needed within a state, we can guarantee that it has been set once we enter that state. An example is the mission ID which is needed in certain states like monitor and stopping. This means that we no longer need to keep this as a global state, and then to hope that it's still set when we reach a location that needs it. Stopping a return home mission passes along the mission we want to start when we succeed at stopping the current return home. Additionally, if we wanted to we could handle more logic within transition functions so that this logic could be moved out of event handlers. Currently I have only made one transition function for each state and have avoided doing any logic within them to keep the example simple, but there is nothing stopping developers from adding more transitions with more logic if they think it's useful. You could for instance have a transition to monitor when you first start a mission, where the transition function itself initiated the process, and then you could have a different transition to monitor which is for when a mission is already running. Besides having a different name, these functions could be differentiated by their parameters. For actions that should be done for all transitions to a state, these can be done in the constructor, which is already done for some states.

You may notice that the type hinting for states in the return values for event handlers is a bit long. This is due to a problem of circular dependencies. To handle this I moved the transitions functions out of the State classes but kept them in the same file. This solves the probelm as we are able to load the transition function without loading the State class, but it does mean that we need to reference the State with its reference in the file as opposed to referencing the class directly (so it is Transitions[Home.Home] instead of Transitions[Home]. The transitions is just Home.transition() though). This doesn't have any functional affect, but it does make the type hinting slightly longer than it would need to be otherwise. If anyone knows any other ways of getting around the circular dependency problem then let me know, but I don't personally mind the current look.

The main feature that is currently lacking is the ability to generate a state transitions diagram. This is theoretically possible, but out of scope for this PR. The state machine class has everything it needs to generate such a diagram, as iterating through the return types for each event handler in a state tells you all the possible state transitions that a state has, though if this ends up feeling clunky then we can always store this information in a separate value in each state that is populated automatically. It is then trivial to use a library like mermaid to generate the state diagram.

I also moved the event handler null checking out of each handler to simplify the code. This was the type hinting of event handlers more accurately represent the possible outcomes of an event.

      There is one new TODO, but it links to a new issue. It can be removed once the issue is addressed.
- [x] A test has been written

### DIFF
--- a/src/isar/state_machine/states/home.py
+++ b/src/isar/state_machine/states/home.py
@@ -106,7 +106,7 @@ class Home(State):
             ),
             EventHandlerMapping(
                 name="stop_mission_event",
-                event=events.api_requests.return_home.request,
+                event=events.api_requests.stop_mission.request,
                 handler=lambda event: stop_mission_event_handler(
                     state_machine, event, None
                 ),


src/isar/state_machine/states/await_next_mission.py
```diff
@@ -107,8 +110,15 @@ def _start_return_home():
         ]
 
         super().__init__(
-            state_name="await_next_mission",
+            state_name=States.AwaitNextMission,
```
andchiind: Note how we know type-check the state names, to make sure they match the enum used when communicating over MQTT.

src/isar/state_machine/utils/common_event_handlers.py
```diff
@@ -11,13 +14,9 @@
...
     state_machine: "StateMachine",
-    event: Event[Mission],
+    mission: Mission,
     response: Event[MissionStartResponse],
-) -> Optional[Callable]:
-    mission: Optional[Mission] = event.consume_event()
-    if not mission:
-        return None
-
+) -> Optional[Transition["Monitor.Monitor"]]:
```
andchiind: Putting the type in quotes here avoids any potential issues with circular depndencies.

tests/isar/state_machine/states/test_await_next_mission_state.py
```diff
@@ -35,10 +42,7 @@ def test_state_machine_with_successful_mission_stop(
         RobotStatusThread, "_is_ready_to_poll_for_status", return_value=True
     )
 
-    # Set the return home delay to a higher value than the test needs to run
-    state_machine_thread.state_machine.await_next_mission_state.timers[
-        0
-    ].timeout_in_seconds = 15
+    mocker.patch.object(settings, "RETURN_HOME_DELAY", 15)
```
andchiind: We now mock the settings instead of actually changing the global variable.

src/isar/state_machine/transitions/mission.py
```diff
@@ -1,225 +0,0 @@
-from typing import TYPE_CHECKING, List
```
Eddasol: Can delete entire folder as well

Eddasol (approved):
LGTM :) This feels more readable

PR #1081, opened 2026-02-09
Add persistent lockdown mode

### LINKED ISSUE #993: State should still be lockdown if ISAR restarts while robot is in lockdown
If the robot is in lockdown, going to lockdown or stopping to go to lockdown, then it should be in the same state when ISAR restarts. This can be achieved with the same functionality used to handle maintenance mode.

If ISAR restarts and it detects that a mission is still ongoing (robot status 'busy') I think it's best to assume that the mission is a lockdown mission. If no mission is running and the robot is localised ('available') then we should start the lockdown mission. If it is 'home' then we can simply go to the lockdown state.

An alternative solution in case the robot is busy is to go to stopping to go to lockdown state. That way we stop the current mission and start a mission we can be 100% sure is a lockdown mission. This might be overly cautious though, since if we are in going to lockdown state and ISAR crashes, then it should be safe to assume that the mission running when we restart is the same one that was running moments before.

It will make lockdown more robust.

N/A

andchiind:
Closes #993 

- [x] A test has been written

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -9,7 +9,6 @@ from typing import Generator
 
 import pytest
 import sqlalchemy
-from dependency_injector.providers import Singleton
 from dependency_injector.wiring import providers
 from fastapi import FastAPI
 from fastapi.testclient import TestClient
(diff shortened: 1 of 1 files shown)


Christdej (approved):
LGTM

PR #1102, opened 2026-03-05
Simplifying robot service logic

andchiind:
The point of this PR is to simplify the robot interface.

Instead of having separate thread objections for each action, we create a generic action thread which runs each event handler. In the eventhandler we both run the action handler (which was previously its own thread) and handle the output, synchronously. Only one action is handled at a time. Other state machine events are ignored in the mean time.

- [x] A test has been written


src/isar/robot/function_thread.py
```diff
@@ -0,0 +1,19 @@
...
+class FunctionThread(Thread):
+    def __init__(
+        self, handler_function: Callable[P, None], *args: P.args, **kwargs: P.kwargs
+    ):
+        self.handler_function = lambda: handler_function(*args, **kwargs)
+        Thread.__init__(self, name="Robot start mission thread")
+        self.start()
+
+    def run(self) -> None:
+        self.handler_function()
```
andchiind: This class just runs whatever function it gets, with whatever arguments it gets. It's useful for when we want a function in its own thread, but we also want to type check its arguments.

tests/isar/services/robot/test_robot_service.py
```diff
@@ -124,75 +114,39 @@ def test_mission_fails_to_stop(
...
-
-    r_service._stop_mission_done_handler()
-
-    assert not r_service.robot_service_events.mission_failed_to_stop.has_event()
-    assert r_service.robot_service_events.mission_successfully_stopped.has_event()
-
-    assert not r_service.robot_service_events.mission_succeeded.has_event()
-    assert not r_service.robot_service_events.mission_failed.has_event()
-
-    assert r_service.monitor_mission_thread is None
```
andchiind: This test became equivalent to the test_successful_stop test after the refactor.

src/isar/robot/function_thread.py
```diff
@@ -0,0 +1,16 @@
...
+
+P = ParamSpec("P")
+
+
+class FunctionThread(Thread):
+    def __init__(
+        self, handler_function: Callable[P, None], *args: P.args, **kwargs: P.kwargs
+    ):
+        self.handler_function = lambda: handler_function(*args, **kwargs)
+        Thread.__init__(self, name="Robot start mission thread")
```
nicholasdalhaug: Change name

nicholasdalhaug (approved):

PR #1113, opened 2026-04-09
Fix: Transition to intervention_needed when battery-low stop fails

### LINKED ISSUE #1112: When the robot fails to stop a mission due to low battery it keeps retrying
When the robot fails to stop a mission when it needs to charge, it currently goes back to monitor from stopping_go_to_recharge. This can lead to loops where it never succeeds to stop the mission. Instead of transitioning to monitor when it fails to stop the mission when running low on battery, it should instead transition to intervention needed so that we can be notified about it (when we enter intervention needed an event is sent on MQTT automatically, so this does not need to be implemented in this PR).

It would allow us to be notified that the robot is unable to return home. There is anyways retry functionality in robot_stop_mission.py, so there is no need to retry stopping in the state machine. If it fails to stop the mission in the state machine then we can reasonably give up and go to intervention needed.

This is not true for the normal, user-initiated stopping functionality, since here it is not an automated system and the user might still want the robot to remain in monitor to see why it failed to stop. The user can then anyways put the robot in maintenance or lockdown if they want to handle the problem.

N/A

asotra:
Fixes #1112

## Problem
When the robot fails to stop a mission due to low battery, it currently transitions back to `monitor` state, creating an infinite retry loop.

## Solution
Changed the state transition to go to `intervention_needed` instead, which automatically sends an MQTT notification to operators.

## Changes
- Modified state transition logic in the state machine
- Failed stops during automated battery-low scenarios now transition to `intervention_needed`
- User-initiated stops still go to `monitor` for debugging purposes

## Testing
- Ran existing test suite
- Verified the state transition change works as expected

- [x] A test has been written

### DIFF
--- a/src/isar/state_machine/states/stopping_go_to_recharge.py
+++ b/src/isar/state_machine/states/stopping_go_to_recharge.py
@@ -1,7 +1,7 @@
 from typing import TYPE_CHECKING, List
 
 import isar.state_machine.states.going_to_recharging as GoingToRecharging
-import isar.state_machine.states.monitor as Monitor
+import isar.state_machine.states.intervention_needed as InterventionNeeded
 from isar.eventhandlers.eventhandler import EventHandlerMapping, State, Transition
 from isar.models.events import EmptyMessage
 from isar.state_machine.states_enum import States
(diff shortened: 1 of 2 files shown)


andchiind (changes requested):
The PR looks good.

Could you just make sure that it passes the linting checks (black, isort, mypy and ruff) for python 3.14? And then could you also change the commit name so that it doesn't start with "fix:"? You could for instance change it to: "Go to intervention needed when battery-low stop fails".

andchiind (approved):
LGTM

PR #1120, opened 2026-04-16
Handle restarting mission after recharging

andchiind:
Closes #1109 

See https://github.com/equinor/flotilla/pull/2678

- [x] A test has been written


src/isar/robot/robot_service.py
```diff
@@ -142,15 +159,16 @@ def _resume_mission_handler(self) -> None:
                 EmptyMessage()
             )
 
-    async def _monitor_mission_handler(self, mission: Mission) -> None:
+    async def _monitor_mission_handler(self, mission: Mission) -> Mission | None:
```
andchiind, line 151: The big change is that the monitor function now returns missions that were not done when exiting monitoring

src/isar/robot/robot_monitor_mission.py
```diff
@@ -162,140 +159,116 @@ def log_task_status(logger: logging.Logger, task: TASKS) -> None:
...
-                        "Failed to collect task status. Error description: %s",
-                        e.error_description,
-                    )
-                    # Currently we only stop mission monitoring after failing to get mission status
-
-            new_mission_status: MissionStatus
             try:
-                new_mission_status = await get_mission_status(robot, mission.id)
+                if mission_status in [
+                    MissionStatus.NotStarted,
```
oysand, line 207: Can this ever be MissionStatus.NotStarted when it is initialized as MissionStatus.InProgress above?

oysand: And if we expect that get_mission_status can return NotStarted, should we initialize it as NotStarted rather than InProgress above?

andchiind: Sure, I can initialize it as notstarted. It doesn't really matter since this function doesn't report mission status anyways.

andchiind: We initialised it like this since initiate_mission was already completed, but that's not an assumption we need to make I suppose.

andchiind: Although since inprogress has already been reported at this point I would prefer it to keep it as is. It also follows the practise of relying on robot service actions for determining mission status, so it would be nice to continue that assumption here.

oysand: Makes sense, then we keep it as is

oysand (approved):
LGTM

PR #1133, opened 2026-05-17
Add analysis_types passthrough from Flotilla to SARA

aeshub:
## Summary

Carry a per-inspection `analysis_types` list from Flotilla through ISAR to SARA, without ISAR inspecting the values. ISAR is a courier here; SARA validates.

- `StartMissionInspectionDefinition.analysis_types: list[str] | None` is threaded onto every `InspectionTask`, copied to `InspectionMetadata`, and re-emitted as `required_analysis` on the MQTT `InspectionResultPayload` consumed by SARA. The list is also written to the metadata sidecar JSON.
- Backward compatible in both directions: missions without the field produce `null` on the wire; SARA falls back to its existing default-analysis-by-file-extension behaviour.
- Wire field uses snake_case to match the surrounding fields in the ISAR-facing serializer on Flotilla's side (`inspection_target`, `inspection_description`).

## Related PRs

- Flotilla: equinor/flotilla#2694 (must add the matching \`analysis_types\` field to \`IsarInspectionDefinition.cs\` to populate this end-to-end)
- SARA: equinor/sara#363 (consumes \`required_analysis\` from the MQTT payload)

## Tests

- New parametrized test in \`test_start_mission_definition.py\` confirms the field defaults to \`None\` for all six inspection types.
- Updated JSON fixture exercises the field end-to-end through \`to_isar_mission\`.
- New uploader tests assert \`required_analysis\` round-trips to the published MQTT JSON for both populated and absent cases.

## Drive-by

- Fixed \`StorageFake.store()\` to return \`StoragePaths\` (previously returned \`None\`, violating the abstract interface) so the new uploader tests can reuse the shared mock.
- Replaced fixed \`time.sleep\` waits in \`tests/isar/storage/test_uploader.py\` with a polling \`_wait_until\` helper where the assertion is on the presence of an event; cuts ~3s off the suite.

### DIFF
--- a/src/isar/apis/models/start_mission_definition.py
+++ b/src/isar/apis/models/start_mission_definition.py
@@ -40,6 +40,7 @@ class StartMissionInspectionDefinition(BaseModel):
     inspection_target: InputPosition
     inspection_description: str | None = None
     duration: float | None = None
+    analysis_types: list[str] | None = Field(default=None)
 
 
 class StartMissionTaskDefinition(BaseModel):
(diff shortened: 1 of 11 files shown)


Christdej (approved):
LGTM

PR #1135, opened 2026-05-17
test: reuse wait helper for async assertions

### LINKED ISSUE #1134: Extend `_wait_until` helper to more tests to reduce flakiness
## Background

PR #1133 introduced a small `_wait_until` polling helper in `tests/isar/storage/test_uploader.py` to replace fixed `time.sleep` calls before assertions. Polling for the awaited condition instead of sleeping a guessed-at duration makes tests more robust to timing variation on CI runners and local machines.

Several other tests in the suite still use fixed sleeps before asserting on state machine transitions or other observable conditions. These are the most common source of flaky test runs.

## Proposal

1. Promote `_wait_until` out of `test_uploader.py` into a shared test utility module so it can be reused.
2. Replace fixed sleeps with `wait_until` in tests where the sleep precedes an assertion on a condition that can be polled (e.g. the state machine reaching a terminal state, a queue containing an expected item, a mock having received a call).
3. Leave sleeps that intentionally assert on *absence* of an event during a window, or that simulate production-like delays in mocks, untouched.

## Why

- **Less flaky tests** — the assertion runs as soon as the condition is true rather than after a worst-case wait, so tests don't fail when CI is slow.
- Faster suite runs as a nice side benefit.

## Good first issue notes

- Scope is mechanical and well-contained: identify candidate sleeps, swap them for `wait_until` with a sensible predicate.
- No production code changes required.
- A good way to get familiar with the state machine tests and the existing test mocks.

tjmyou123:
Fixes #1134

## Summary
- Add a shared test wait helper for polling asynchronous conditions.
- Replace selected fixed sleeps in uploader and state-machine tests with condition-based waits.
- Leave absence-check sleeps in place where the wait window is part of the assertion.

## Testing
- Syntax checked changed files with Pylance: no syntax errors.
- Targeted pytest command was attempted but blocked by an existing syntax error in src/robot_interface/telemetry/mqtt_client.py:113 under the local Python 3.11 environment.

### DIFF
--- a/tests/isar/state_machine/states/test_await_next_mission_state.py
+++ b/tests/isar/state_machine/states/test_await_next_mission_state.py
@@ -1,4 +1,3 @@
-import time
 from collections import deque
 from typing import cast
 
@@ -27,6 +26,7 @@ from tests.test_mocks.state_machine_mocks import (
     UploaderThreadMock,
 )
 from tests.test_mocks.task import StubTask
+from tests.test_utils import wait_until
 
 
 def test_state_machine_with_successful_mission_stop(
@@ -56,13 +56,17 @@ def test_state_machine_with_successful_mission_stop(
     state_machine_thread.start()
     robot_service_thread.start()
     uploader_thread.start()
-    time.sleep(1)
+    assert wait_until(
+        lambda: state_machine_thread.state_machine.transitions_list
+        == deque([States.UnknownStatus, States.Home])
+    )
     scheduling_utilities.start_mission(mission=mission)
-    time.sleep(0.5)
+    assert wait_until(
+        lambda: state_machine_thread.state_machine.current_state.name == States.Monitor
+    )
     scheduling_utilities.stop_mission(mission_id=mission.id)
-    time.sleep(3)  # Allow enough time to stop the mission
 
-    assert state_machine_thread.state_machine.transitions_list == deque(
+    expected_transitions = deque(
         [
             States.UnknownStatus,
             States.Home,
@@ -71,6 +75,11 @@ def test_state_machine_with_successful_mission_stop(
             States.AwaitNextMission,
         ]
     )
+    assert wait_until(
+        lambda: state_machine_thread.state_machine.transitions_list
+        == expected_transitions
+    )
+    assert state_machine_thread.state_machine.transitions_list == expected_transitions
 
 
 def test_transition_from_resuming_to_paused(
--- a/tests/isar/state_machine/states/test_blocked_protective_stop_state.py
+++ b/tests/isar/state_machine/states/test_blocked_protective_stop_state.py
@@ -1,4 +1,3 @@
-import time
 from collections import deque
 
 from pytest_mock import MockerFixture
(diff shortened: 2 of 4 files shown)


tests/isar/state_machine/states/test_blocked_protective_stop_state.py
```diff
@@ -29,8 +29,12 @@ def test_state_machine_idle_to_blocked_protective_stop_to_idle(
...
 
-    assert state_machine_thread.state_machine.transitions_list == deque(
+    expected_transitions = deque(
         [States.UnknownStatus, States.BlockedProtectiveStop, States.Home]
     )
+    assert wait_until(
+        lambda: state_machine_thread.state_machine.transitions_list
+        == expected_transitions
+    )
+    assert state_machine_thread.state_machine.transitions_list == expected_transitions
```
aeshub, line 40: This assert is not required as the line above effectively already asserts the same thing. Please remove it.

tests/isar/state_machine/states/test_await_next_mission_state.py
```diff
@@ -71,6 +75,11 @@ def test_state_machine_with_successful_mission_stop(
             States.AwaitNextMission,
         ]
     )
+    assert wait_until(
+        lambda: state_machine_thread.state_machine.transitions_list
+        == expected_transitions
+    )
+    assert state_machine_thread.state_machine.transitions_list == expected_transitions
```
aeshub, line 82: This assert is not required as the line above effectively already asserts the same thing. Please remove it.

tests/isar/storage/test_uploader.py
```diff
@@ -67,9 +68,9 @@ def test_should_upload_from_queue(
     uploader: Uploader = container.uploader()
 
     uploader.upload_queue.put(message)
-    time.sleep(0.01)
 
     storage_handler: StorageFake = uploader.storage_handlers[0]  # type: ignore
+    assert wait_until(lambda: inspection in storage_handler.stored_inspections)
     assert inspection in storage_handler.stored_inspections
```
aeshub, line 74: This assert is not required as the line above effectively already asserts the same thing. Please remove it.

tests/isar/storage/test_uploader.py
```diff
@@ -98,9 +99,9 @@ def test_should_retry_failed_upload_from_queue(
     # Should not upload, instead raise StorageException
     assert not storage_handler.blob_exists(inspection)
     storage_handler.will_fail = False
-    time.sleep(3)
 
     # After some time, it should have retried and now it should be successful
+    assert wait_until(lambda: storage_handler.blob_exists(inspection))
     assert storage_handler.blob_exists(inspection)
```
aeshub, line 105: This assert is not required as the line above effectively already asserts the same thing. Please remove it.

tests/isar/storage/test_uploader.py
```diff
@@ -127,8 +128,8 @@ def test_should_not_publish_when_blob_paths_are_empty(
         mission,
     )
     uploader.upload_queue.put(message)
-    time.sleep(1)
 
+    assert wait_until(lambda: inspection in storage_handler.stored)
     assert inspection in storage_handler.stored
```
aeshub, line 133: This assert is not required as the line above effectively already asserts the same thing. Please remove it.

aeshub (changes requested):
Thank you for the contribution! Please have a look at the comments provided.

Additionally, we do not use conventional commits here so please adjust the commit message. For example, it could be "Reuse wait helper for async assertions" instead.

aeshub:
Closing due to inactivity.

PR #1144, opened 2026-06-29
Refactor thread sleep intervals slightly

andchiind:
Use thread.sleep instead of constantly checking a variable, which makes the threads more consistent.

Increase time between info publishers to avoid unnecessary load on listeners. It is not an issue if it takes slightly longer to be notified of new robots.

- [ ] A test has been written
  - [x] This change doesn't need a new test (tests have been updated)


Christdej (approved):
LGTM

PR #1145, opened 2026-07-06
Smaller state machine refactors

### LINKED ISSUE #808: Cannot know if robot is both home and protective stop
**Describe the bug**
Currently both home and protective_stop are potential options for robot_status, but what if robot is both home and protective_stop is active. Should it say home or protective stop? This is up to the robot package but should have information of both these things

**To Reproduce**

**Expected behavior**

**Screenshots**

andchiind:
Breaking change since I renamed one of the robot status names in the second commit.

- [x] A test has been written


src/isar/state_machine/states/maintenance.py
```diff
@@ -59,6 +52,10 @@ def _release_from_maintenance_handler(
 
 def transition() -> Transition[Maintenance]:
     def _transition(state_machine: "StateMachine") -> Maintenance:
+        state_machine.events.api_requests.set_maintenance_mode.response.trigger_event(
+            MaintenanceResponse(is_maintenance_mode=True)
+        )
```
marianards, line 57: I think here we might have issues, Maintenance.transition() is now also called from the teleoperation path in home.py:60 and unknown_status.py and they are not initiated by a set_maintenance_mode API request. 

This will fire a set_maintenance_mode.response with no caller and a stale MaintenanceResponse can sit in the queue and be picked up by the next real API call, making it return success immediately without actually doing the request round-trip

marianards: You might have to do the same thing as in transition_and_respond_to_api and transition_without_responding_to_api

src/isar/state_machine/states/unknown_status.py
```diff
@@ -21,29 +19,22 @@ class UnknownStatus(State):
     def __init__(self, state_machine: "StateMachine"):
         # Ensures that we will check the status immediately instead of waiting for it to change
```
marianards: stale comment

src/isar/state_machine/states/maintenance.py
andchiind: Good point

src/isar/state_machine/states/unknown_status.py
andchiind: Good catch

marianards (approved):
LGTM :)

PR #1146, opened 2026-07-13
More state machine simplifications

andchiind:
The aim is to work towards a functional state machine without side effects within eventhandlers. It is nice to restrict side-effects to transition functions, so that the caller is aware of what it will do (assuming the transition has a nice name) and so that the state being transitioned to is able to maintain its own logic without needing to worry about what happened before the transition (eg, a state that should have a mission running in it needs to be able to know that the mission has been initiated).

Remaining side effects within eventhandlers:
- publishing mission status on MQTT
- replying to stop mission API requests
- replying to release from maintenance/lockdown API requests
- checking if the stop mission request ID matches the current ongoing mission ID (arguably not a side effect)
- checking how many times we've retried return home (arguably not a side effect)

If we got rid of all the side effects then we wouldn't need to pass the state machine reference to the different States, which would prevent side effects. All they would need would instead be just the events.

- [x] A test has been written


src/robot_interface/models/mission/mission.py
```diff
@@ -36,3 +36,8 @@ def _get_unfinished_tasks(self) -> List[TASKS]:
                 self.tasks,
             )
         )
+
+
+class ReturnHomeMission(Mission):
+    tasks: List[TASKS] = [ReturnToHome()]
```
marianards:
```suggestion
    tasks: List[TASKS] = Field(default_factory=lambda: [ReturnToHome()])
```

The old code created a fresh task each call, but now the ReturnToHome() is constructed once at class-definition time, so i believe every ReturnHomeMission() is reusing the same task Id

src/isar/robot/robot_battery.py
```diff
@@ -30,22 +40,52 @@ def run(self) -> None:
...
-                self.shared_state.robot_battery_level.update(robot_battery_level)
+                if (
+                    battery_level > settings.ROBOT_BATTERY_RECHARGE_THRESHOLD
+                    and last_battery_value < settings.ROBOT_BATTERY_RECHARGE_THRESHOLD
+                ):
+                    self.battery_above_recharge_threshold_event.trigger_event(
+                        EmptyMessage()
+                    )
+                else:
+                    self.battery_above_recharge_threshold_event.clear_event()
```
marianards, line 65: Optional: Looks like this event is only "set" for a poll interval, for the other mission-threshold events we stay set until the opposite, which is more robust. Could consider doing the same here

src/isar/robot/robot_battery.py
```diff
@@ -30,22 +40,52 @@ def run(self) -> None:
...
+
+                if (
+                    battery_level < settings.ROBOT_MISSION_BATTERY_START_THRESHOLD
+                    and last_battery_value
+                    > settings.ROBOT_MISSION_BATTERY_START_THRESHOLD
+                ):
+                    self.battery_above_mission_threshold_event.clear_event()
+                    self.battery_below_mission_threshold_event.trigger_event(
```
marianards: I think no state is consuming battery_above_mission_threshold_event so its dead code

src/isar/robot/robot_battery.py
andchiind: Good point

src/isar/robot/robot_battery.py
andchiind: Good point

marianards (approved):
LGTM :D

PR #1155, opened 2026-07-30
Simplify uploader and make tests deterministic

andchiind:
There are two ways to receive inspection data, either by querying the robot, or via a callback function provided on the robot interface.

For the former, we would create an event when we were monitoring the mission and saw that a task completed. This event would then be read in the robot inspection service thread, which would create one thread per such request, which would fetch the inspection data from the robot interface, then put it on the queue which the uploader thread used. The uploader thread would then check this queue, wrap the data in some temporary data structure, then add it to another queue which it would check constantly within the same thread, retrying to upload as needed.

For the latter, the callback function would be registered within its own thread within the robot inspection service. In that thread the output of the callback function would be placed directly on the queue used by the uploader thread.

In this new system we keep the robot inspection service, but get rid of the uploader thread. The uploader is kept as a module, since it is convenient to be able to do dependency injection, but it can be used as an uploader function with some nice wrappers. When we get an event about a task completing when monitoring the mission, the robot service thread creates one thread which reads the inspection data from the robot interface, and creates an inspection data event. The callback function for the async inspection data creates this inspection data event directly. The robot inspection service also reads these events, and creates one thread per such event which does the same thing the uploader thread used to do, but for one inspection event at a time.

This not only simplifies the number of thread types we have, and the number of places they are created, but also simplifies the uploader module in such a way that we can test it synchronously, in deterministic unit tests.

- [x] A test has been written

### DIFF
--- a/src/isar/models/events.py
+++ b/src/isar/models/events.py
@@ -16,8 +16,6 @@ from robot_interface.models.mission.status import RobotStatus
 from robot_interface.models.mission.task import InspectionTask
 from robot_interface.telemetry.mqtt_client import MQTTQueueType
 
-InspectionQueueTuple = tuple[Inspection, Mission]
-
 
 class EmptyMessage:
     def __str__(self) -> str:
@@ -28,8 +26,8 @@ AbortedMission = Mission
 
 
 class Event[T](Queue[T]):
-    def __init__(self, name: str) -> None:
-        super().__init__(maxsize=1)
+    def __init__(self, name: str, maxsize: int = 1) -> None:
+        super().__init__(maxsize=maxsize)
         self.name = name
 
     def trigger_event(self, data: T, timeout: int | None = None) -> None:
@@ -90,8 +88,8 @@ class Events:
         self.state_machine_events: StateMachineEvents = StateMachineEvents()
         self.robot_service_events: RobotServiceEvents = RobotServiceEvents()
 
-        self.upload_queue: Queue[InspectionQueueTuple] = Queue[InspectionQueueTuple](
-            maxsize=10
+        self.upload_event: Event[tuple[Inspection, Mission]] = Event(
+            "uploader", maxsize=10
         )
 
         self.mqtt_queue: Queue[MQTTQueueType] = Queue[MQTTQueueType]()
--- a/src/isar/robot/robot_inspection_service.py
+++ b/src/isar/robot/robot_inspection_service.py
@@ -1,18 +1,12 @@
 import logging
 from collections.abc import Callable
-from queue import Queue
 from threading import Event as ThreadEvent
 from threading import Thread
 
 from isar.config.settings import settings
-from isar.models.events import (
-    EventConflictError,
-    Events,
-    EventTimeoutError,
-    RobotServiceEvents,
-    StateMachineEvents,
-)
+from isar.models.events import Event, EventConflictError, Events, EventTimeoutError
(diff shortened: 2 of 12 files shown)


src/isar/storage/uploader.py
```diff
@@ -36,254 +32,153 @@ def has_empty_blob_storage_path(storage_paths: StoragePaths) -> bool:
...
-            target_position=inspection.metadata.target_position,
-        )
-        self.mqtt_publisher.publish(
-            topic=settings.TOPIC_ISAR_INSPECTION_RESULT,
-            payload=payload.model_dump_json(),
-            qos=1,
-            retain=True,
-            properties=props_expiry(settings.MQTT_MISSION_TASK_AND_STATUS_EXPIRY),
+                time.sleep(sleep_length)
+        logger.error(
```
marianards, line 121: should this not be idented at the while loop instead of inside? i think now each attempt is logging exceeded max retries

andchiind: Good catch

src/isar/robot/robot_inspection_service.py
```diff
@@ -56,7 +49,7 @@ def robot_upload_inspection(
         inspection,
         mission,
     )
-    upload_queue.put(message)
+    upload_inspection_event.trigger_event(message)
```
nicholasdalhaug: I cannot see why the `fecth_and_upload_inspection` is not running the `uploader.upload_inspection`

src/isar/models/events.py
```diff
@@ -90,8 +88,8 @@ def __init__(self) -> None:
         self.state_machine_events: StateMachineEvents = StateMachineEvents()
         self.robot_service_events: RobotServiceEvents = RobotServiceEvents()
 
-        self.upload_queue: Queue[InspectionQueueTuple] = Queue[InspectionQueueTuple](
-            maxsize=10
+        self.upload_queue: Event[tuple[Inspection, Mission]] = Event(
```
nicholasdalhaug: Should this not be a Queue and not an event? Later you have written "upload_queue.trigger_event" which is confusing. It is mostly just a "put" anyways, but if this had been a Queue and not an Event, the difference would be more understandable.

nicholasdalhaug (commented):
Looks very good. I add the comments instead of requesting changes so they are not required. I would however like to understand more.

src/isar/models/events.py
andchiind: I can rename it to something other than upload_queue. I wanted to use the existing Event class so that it would have the same error handling and behaviour as the other events, but it's not very important tbh. It was mainly so that we didn't have a mix of events and native queues, but there's almost no difference.

src/isar/robot/robot_inspection_service.py
andchiind: I did it this way to have a single location where all the upload events were handled, but you're right that there would be the same number of threads either way. I don't have strong opinions on it, I can just call it here like you mention

Eddasol (approved):
LGTM

src/isar/models/events.py
andchiind: I have updated the name. I tried using a Queue again, but it's really convenient to have the error handling in trigger_event and consume_event.

src/isar/robot/robot_inspection_service.py
andchiind: I have updated it as suggested, to avoid the extra step

PR #1157, opened 2026-07-31
Auto test valid API states

andchiind:
Instead of relying on people manually setting which state the robot can accept API requests in, this PR does it automatically and tests based on a specification.

- [x] A test has been written

### DIFF
--- a/src/isar/state_machine/state.py
+++ b/src/isar/state_machine/state.py
@@ -64,6 +64,10 @@ class State(ABC):
         )
         return filtered_timers[0] if len(filtered_timers) > 0 else None
 
+    def handles_event(self, event: Event) -> bool:
+        allowed_events: list[Event] = [m.event for m in self.event_handler_mappings]
+        return event in allowed_events
+
     def run(self) -> Transition | None:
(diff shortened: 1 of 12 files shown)


src/isar/state_machine/state.py
```diff
@@ -64,6 +64,9 @@ def get_event_timer_by_name(
         )
         return filtered_timers[0] if len(filtered_timers) > 0 else None
 
+    def handles_event(self, event: Event) -> bool:
+        return event in [m.event for m in self.event_handler_mappings]
```
olaals: Nitpicky, but could this be a variable assignment in its own line to make the variable name be a bit documenting?

Probably nice and compact this way but it would be nice with either a comment or a variable name so it is more easily understood

allowed_state_events = [m.event for m in self.event_handler_mappings]
return event in allowed_state_events

olaals (approved):
LGTM

src/isar/state_machine/state.py
andchiind: Probably a good idea. Something I am maybe a bit too functional 😬

PR #1158, opened 2026-08-05
Support non-Azure OpenID providers

oysand:
Adds `ISAR_OPENID_CONFIG_URL`, which points token validation at any OpenID Connect provider instead of Azure Entra ID — a Keycloak realm for local development and for the armada integration tests, or an operator's own issuer for a deployment outside Azure. Unset, which is the default, nothing changes.

Issuer validation stays enabled either way; the expected issuer comes from the discovery document. `ISAR_OPENID_SCOPE` and the two `ISAR_OPENAPI_*` variables keep Swagger's Authorize button off `login.microsoftonline.com`. The scheme is extracted into a factory so it can be tested without reloading the module.

Also asserts that `aud` must be a string, not the array form RFC 7519 permits, since `fastapi-azure-auth` rejects the latter with an opaque 401.

Verified: 165/165 tests, mypy and ruff clean, `uv.lock` unmoved. Exercised end to end in the full integration suite (equinor/armada#87, 4/4 green) with an `isar-robot` image built from this branch.

Merge order: **this PR** → isar-robot bump → equinor/flotilla#2863 + equinor/sara#456 → images published → equinor/armada#87.

Supersedes #1156, which stays open as a fallback until this set is approved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,21 +270,46 @@ option in the dictionary.
 
 ## API authentication
 
-The API has an option to include user authentication. This can be enabled by setting the environment variable
+The API validates OAuth2 access tokens. Authentication is controlled by
 
 ```
 ISAR_AUTHENTICATION_ENABLED = true
 ```
 
-By default, the `local` storage module is used and API authentication is disabled. If using Azure Blob Storage a set of
-environment variables must be available which gives access to an app registration that may use the storage account.
-Enabling API authentication also requires the same environment variables. The required variables are
+which is **enabled by default**; set it to `false` to turn authentication off.
+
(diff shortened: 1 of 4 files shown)


oysand:
Verified end to end in the full integration suite (equinor/armada#87, green), using an `isar-robot` image built from this branch. ISAR validated Keycloak-issued tokens throughout, the fixture invariant confirmed `POST /schedule/stop-mission` refuses unauthenticated callers, and the woven negative assertions confirmed ISAR rejects both a wrong-audience token and one missing `Mission.Control` while a mission was in flight.

Christdej (approved):
LGTM

PR #1159, opened 2026-08-06
Adapt to ruff recommendations

Eddasol:
- [ ] A test has been written

### DIFF
--- a/src/isar/models/events.py
+++ b/src/isar/models/events.py
@@ -1,5 +1,5 @@
 from collections import deque
-from queue import Empty, Queue
+from queue import Empty, Full, Queue, ShutDown
 from threading import Lock
 from typing import Generic, TypeVar
 
@@ -43,7 +43,7 @@ class Event(Queue[T]):
             # We always want a timeout when blocking for results, so that
             # the thread will never get stuck waiting for a result
             self.put(data, block=timeout is not None, timeout=timeout)
-        except Exception:
+        except Full, ShutDown, ValueError:
             if timeout is not None:
                 raise EventTimeoutError
             return
@@ -55,7 +55,7 @@ class Event(Queue[T]):
             if timeout is not None:
                 raise EventTimeoutError
             return None
-        except ValueError:
+        except ValueError, ShutDown:
             raise EventConflictError
 
     def clear_event(self) -> bool:
--- a/src/isar/models/events.py
+++ b/src/isar/models/events.py
@@ -46,7 +46,7 @@ class Event(Queue[T]):
         except Exception:
             if timeout is not None:
                 raise EventTimeoutError
-            return None
+            return
 
     def consume_event(self, timeout: int | None = None) -> T | None:
         try:
--- a/src/isar/models/events.py
+++ b/src/isar/models/events.py
@@ -1,7 +1,6 @@
 from collections import deque
 from queue import Empty, Full, Queue, ShutDown
 from threading import Lock
-from typing import Generic, TypeVar
 
 from isar.apis.models.models import (
     ControlMissionResponse,
@@ -17,11 +16,6 @@ from robot_interface.models.mission.status import RobotStatus
 from robot_interface.models.mission.task import InspectionTask
 from robot_interface.telemetry.mqtt_client import MQTTQueueType
 
-T = TypeVar("T")
-T1 = TypeVar("T1")
-T2 = TypeVar("T2")
-
-
 InspectionQueueTuple = tuple[Inspection, Mission]
 
 
@@ -33,7 +27,7 @@ class EmptyMessage:
 AbortedMission = Mission
 
 
-class Event(Queue[T]):
+class Event[T](Queue[T]):
     def __init__(self, name: str) -> None:
         super().__init__(maxsize=1)
         self.name = name
@@ -105,7 +99,7 @@ class Events:
         self.state: Event[States] = Event("state")
 
 
-class APIEvent(Generic[T1, T2]):
+class APIEvent[T1, T2]:
     """
     Creates request and response event. The events are defined such that the request is from
     api to state machine while the response is from state machine to api.
--- a/src/isar/models/events.py
+++ b/src/isar/models/events.py
@@ -1,7 +1,7 @@
 from collections import deque
 from queue import Empty, Queue
 from threading import Lock
-from typing import Generic, Tuple, TypeVar
+from typing import Generic, TypeVar
 
 from isar.apis.models.models import (
     ControlMissionResponse,
@@ -22,7 +22,7 @@ T1 = TypeVar("T1")
 T2 = TypeVar("T2")
 
 
-InspectionQueueTuple = Tuple[Inspection, Mission]
+InspectionQueueTuple = tuple[Inspection, Mission]
 
 
 class EmptyMessage:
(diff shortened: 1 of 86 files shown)


src/isar/models/events.py
```diff
@@ -43,7 +43,7 @@ def trigger_event(self, data: T, timeout: int | None = None) -> None:
             # We always want a timeout when blocking for results, so that
             # the thread will never get stuck waiting for a result
             self.put(data, block=timeout is not None, timeout=timeout)
-        except Exception:
+        except Full, ShutDown, ValueError:
```
marianards, line 40: except (Full, ShutDown, ValueError):

src/isar/models/events.py
marianards: Should also handle ShutDown to keep API symmetric with trigger_event

except (ValueError, ShutDown):
            raise EventConflictError

src/isar/robot/robot_battery.py
```diff
@@ -85,9 +85,7 @@ def run(self) -> None:
             except RobotException as e:
                 self.logger.error(f"Failed to retrieve robot battery level: {e}")
                 continue
-            except Exception as e:
-                self.logger.error(
-                    f"Unhandled exception in robot battery service: {str(e)}"
-                )
+            except EventTimeoutError as e:
+                self.logger.error(f"An error with the event queue occurred: {str(e)}")
```
marianards: trigger_event is called without a timeout so this will never fire, you can replace it with this:

except Exception as e:  # noqa: BLE001
                self.logger.error(
                    f"Unexpected error in robot battery service: {str(e)}"
                )
                continue

or you can remove the dead code

src/isar/robot/robot_status.py
```diff
@@ -53,9 +53,9 @@ def run(self) -> None:
             except RobotException as e:
                 self.logger.error(f"Failed to retrieve robot status: {e}")
                 continue
-            except Exception as e:
+            except EventTimeoutError as e:
                 self.logger.error(
-                    f"Unhandled exception in robot status service: {str(e)}"
+                    f"Unhandled error occurred with the event queue in robot status service: {str(e)}"
```
marianards: Same comment as for robot_battery.py

src/isar/robot/robot_battery.py
Eddasol: I would prefer to not add except Exception as e:. Is it sufficient to just remove the except EventTimeoutError?

src/isar/services/utilities/scheduling_utilities.py
marianards: except (EventTimeoutError, EventConflictError):

src/isar/robot/robot_pause_mission.py
```diff
@@ -40,14 +40,6 @@ def robot_pause_mission(
...
-            logger.error(
-                f"\nAn unexpected error occurred while pausing the robot: {str(e)}"
-            )
-            return ErrorMessage(
-                error_reason=ErrorReason.RobotUnknownErrorException,
-                error_description=str(e),
-            )
         return None
 
     assert error is not None
```
marianards, line 45: I would refactor this: in theory we are catching all errors and therefore error will never be none, however for future changes of the code this could slip, so having it a bit more reduntant could be good

if error is None:
        return ErrorMessage(
            error_reason=ErrorReason.RobotActionException,
            error_description=f"Failed to pause the robot after {retries} attempts",
        )

It is just a suggestion, you can leave it as is 😄

marianards: same for files robot_resume_mission.py anf robot_stop_mission.py

Eddasol: PR is already massive. Can create issue on this

src/isar/models/events.py
Eddasol: image
Tried to change but black reverted it

src/isar/robot/robot_battery.py
marianards: Yep, it is, you can just keep the RobotException

src/isar/robot/robot_inspection_service.py
marianards: try:
            self.inspection_callback_thread.start()
        except RuntimeError as e:
            self.logger.error(f"Could not restart inspection callback thread: {e}")

The restart_inspection_thread_if_stopped method calls .start() on a dead thread in other bits of the code, but a Thread can only be started once, so it always raises RuntimeError. This was not an issue before because we caught general exceptions, but when keeping it event errors only, that RuntimeError now needs to be catched. If we fix it here then we dont need to worry about all the bits of the code that uses it

src/isar/robot/robot_service.py
```diff
@@ -242,8 +244,8 @@ async def _run_main_event_loop(self) -> None:
 
                 await asyncio.sleep(0)
 
-        except Exception as e:
-            self.logger.error(f"Unhandled exception in robot service: {str(e)}")
+        except (EventTimeoutError, EventConflictError) as e:
+            self.logger.error(f"An error occurred with the event queue: {str(e)}")
```
marianards: This is also dead code, can be removed for the same reason as the others, we only need RobotException

src/isar/models/events.py
marianards: On Python 3.14, except Full, ShutDown, ValueError without parentheses is valid and catches all three. It is not a syntax error, so no fix is needed, sorry about that

src/isar/robot/robot_service.py
Eddasol: We cant have a try without any excepts. Should I remove the entire try/except?

src/isar/robot/robot_pause_mission.py
Eddasol: https://github.com/equinor/isar/issues/1160
Add more details if you have

src/isar/robot/robot_inspection_service.py
Eddasol: Can add, but this is scope creep and not really part of what the PR intends to do

src/isar/robot/robot_service.py
marianards: You can leave the EventConflictError, since we have added ShutDown to the consume_event 
except EventConflictError as e:
            self.logger.error(f"An error occurred with the event queue: {str(e)}")
        self.logger.info("Exiting robot service main thread")

src/isar/robot/robot_inspection_service.py
marianards: This problem has been introduced by your changes. Before, the general except Exception would catch this, but now it doesnt unless you include the runtime error. Not fixing this risks inspection uploads to stop permanently

tests/test_mocks/robot_interface.py
```diff
@@ -19,18 +19,21 @@
 from robot_interface.robot_interface import RobotInterface
 from tests.test_mocks.inspection import stub_image_metadata
 
+_ROBOT_FRAME = Frame(name="robot")
+_DEFAULT_POSITION = Position(x=0, y=0, z=0, frame=_ROBOT_FRAME)
+_DEFAULT_ORIENTATION = Orientation(x=0, y=0, z=0, w=1, frame=_ROBOT_FRAME)
+_DEFAULT_POSE = Pose(
+    position=_DEFAULT_POSITION, orientation=_DEFAULT_ORIENTATION, frame=_ROBOT_FRAME
+)
+
```
andchiind, line 28: Is this related to the ruff update? I don't mind it, just curious

andchiind (approved):
LGTM

tests/test_mocks/robot_interface.py
Eddasol: Yes, the recommendation is to not have function calls inside the function arguments

Eddasol: https://docs.astral.sh/ruff/rules/function-call-in-default-argument/

PR #1164, opened 2026-08-18
Mqtt refactor

### LINKED ISSUE #1163: Refactor MqttPublisher
Currently this class serves as a somewhat convoluted wrapper around the mqtt_queue. I dno't suggest we remove the functionality, but make it into a utility function as opposed to a module. This would also make it more accessible, since not all parts of the code can conveniently do dependency injection without significant changes.

It will make it easier to publish values to MQTT without the need for dependency injection.

N/A

andchiind:
Closes #1163 

Previously all telemetry thread logic was done within each isar implementation. This PR moves the responsibility to ISAR, reducing the overall complexity in the robot interface, whilst providing greater control/error handling within the telemetry threads. In some cases the telemetry thread functionality was hackily used to implement general long lived threads for other purposes in ISAR instances. For this purpose I have instead created a new robot interface function for starting generic threads.

There are several commits that change the robot interface, but the hope is that all commits in this PR will be included in a single ISAR release so that we only need to update the ISAR implementations once.

See:
- https://github.com/equinor/isar-robot/pull/396
- https://github.com/equinor/isar-anymal/pull/27
- https://github.com/equinor/isar-taurob/pull/806

- [ ] A test has been written


src/robot_interface/telemetry/mqtt_client.py
```diff
@@ -1,162 +1,87 @@
...
             except RobotTelemetryPoseException, RobotTelemetryNoUpdateException:
-                time.sleep(self.interval)
                 continue
             except RobotTelemetryException:
                 payload = json.dumps(
                     CloudHealthPayload(
                         isar_id=isar_id,
                         robot_name=robot_name,
-                        timestamp=datetime.now(UTC),
+                        timestamp=str(datetime.now(UTC)),
```
Christdej, line 71: Doesnt this expect a datetime and not a str?

src/isar/models/mqtt_queue.py
```diff
@@ -0,0 +1,184 @@
...
+        queue_message: MQTTQueueMessage = MQTTQueueMessage(
+            topic=topic,
+            payload=payload,
+            qos=qos,
+            retain=retain,
+            properties=properties,
+        )
+        try:
+            self.queue.put(queue_message, timeout=10)
+        except Full:
```
Christdej, line 69: FYI, this will discard all pending messages, and replacing the whole queue on full might loose still valid backlog, consider removing oldest messages and logging which ones are removed

src/isar/models/mqtt_queue.py
```diff
@@ -0,0 +1,184 @@
...
+            self.logger.error(
+                "MQTT queue filled up before being emptied. Messages lost"
+            )
+            self.queue = Queue(maxsize=self.queue.maxsize)
+
+    def get(self, timeout: int = 0) -> MQTTQueueMessage:
+        try:
+            return self.queue.get(block=timeout is not None, timeout=timeout)
```
Christdej: timeout is not None is always true as the parameter for the get function has a inital not non value

src/isar/models/mqtt_queue.py
Christdej, line 1: Consider adding a test for this file

Christdej (changes requested):

src/robot_interface/telemetry/mqtt_client.py
andchiind: You get an exception if you use a datetime directly, since you can't automatically convert it to a JSON. It is anyways a string when it is sent.

src/isar/models/mqtt_queue.py
andchiind: That's true, I guess I can reduce the amount of lost messages that way. The most important thing is that it is logged. Previously we had no error handling for this, and we just got lucky that it never happened in practise, but I was able to test it.

src/isar/models/mqtt_queue.py
andchiind: Its functionality is included in the integration tests already. I removed the mock that was used before so now it is actually tested.

andchiind: Example: https://github.com/equinor/isar/pull/1164/changes#diff-d8b9247c2380726fe34f25a9efba13fa77f4a83d4c0f0eb8260d7d81cbe6adc6

src/robot_interface/telemetry/mqtt_client.py
andchiind: In fact, the code that was there before for handling cloud health messages didn't actually work. With this change it does.

src/isar/models/mqtt_queue.py
andchiind: Sure I can change it to "timeout is not 0"

src/isar/models/mqtt_queue.py
andchiind: And here is the mock I deleted https://github.com/equinor/isar/pull/1164/changes#diff-be3e9176cc8fd520abb981e8c1eb0aee38ab12975c90b9d58e867c079e41349b

Christdej (approved):
LGTM

PR #1167, opened 2026-08-25
Ensure robot reports failure to return home

andchiind:
The logic is that we don't want the possibility that the robot is stuck in paused return home without being able to continue, due to the isar implementation being unable to stop the mission. So in this case we go into intervention needed so that the issue can be reported and addressed.

- [ ] A test has been written
  - [x] This change doesn't need a new test (tests updated)


oysand (approved):
LGTM

PR #1174, opened 2026-09-02
Handle going to lockdown with mission

andchiind:
- [x] A test has been written


src/isar/state_machine/state_machine.py
Christdej, line 104: If you add new states that should be persistent, you need to add them to this list, or they will not be stored

src/isar/state_machine/states/going_to_lockdown_with_mission.py
```diff
@@ -0,0 +1,80 @@
...
+        ),
+        EventHandlerMapping[EmptyMessage](
+            event=events.robot_service_events.mission_failed_to_resume,
+            handler=_lockdown_mission_failed_to_resume,
+        ),
+        EventHandlerMapping[EmptyMessage](
+            event=events.robot_service_events.mission_succeeded,
+            handler=lambda _: LockdownWithMission.transition(mission),
+        ),
+    ]
```
Christdej, line 45: GoignToRechargingWithMission still routes send_to_lockdown to transition_to_existing_mission_and_report_to_api(), since this PR removes the publish_mission_aborted call, mission might stay as ongoing in Flotilla forever

Christdej (changes requested):

src/isar/state_machine/states/going_to_lockdown_with_mission.py
andchiind: Not reporting the mission as aborted is the point of the PR. From Flotilla's POV the mission is just temporarily paused until the robot returns from lockdown.

src/isar/state_machine/state_machine.py
andchiind: Good point, I will add it.

src/isar/state_machine/states/going_to_lockdown_with_mission.py
andchiind: Unless I misunderstand.

Christdej: Agreed that suppressing the abort is the point, and that works for the StoppingGoToLockdown path since GoingToLockdownWithMission/LockdownWithMission carry the mission and resume it. The concern is specifically going_to_recharging_with_mission.py:45: GoingToLockdown.transition_to_existing_mission_and_report_to_api() takes no mission, so the mission is dropped and release from Lockdown goes to Home, not Monitor. Nothing is published either, so Flotilla sees Ongoing forever. Pre-existing, so fine as a follow-up — but the new going_to_lockdown_with_mission.transition_to_existing_mission_and_report_to_api(mission) you added is currently unused and looks like it was written for exactly this. Same for recharging_with_mission.py:37.

Christdej: Also Flotilla might render a mission as "Ongoing" for the entire lockdown duration, could publish MissionStatus.Paused on entry to LockdownWithMission

andchiind: Good point. I will make sure that it transitions to goingtolockdownwithmission in going_to_recharging_with_mission

Christdej (approved):
LGTM

PR #1176, opened 2026-09-03
Refactor state machine events to request response events

andchiind:
This PR not only simplifies the events a bit, but also ensures that related events are cleared before a new result is expected. Previously this was done ad-hoc, leaving it prone to bugs.

- [ ] A test has been written

### DIFF
--- a/src/isar/state_machine/states/going_to_lockdown_with_mission.py
+++ b/src/isar/state_machine/states/going_to_lockdown_with_mission.py
@@ -31,15 +31,15 @@ def GoingToLockdownWithMission(events: Events, mission: AbortedMission) -> State
 
     event_handlers: list[EventHandlerMapping] = [
         EventHandlerMapping[ErrorMessage](
-            event=events.robot_service_events.mission_failed,
+            event=events.action_requests.execute_mission.failure,
             handler=_lockdown_mission_failed,
         ),
         EventHandlerMapping[EmptyMessage](
-            event=events.robot_service_events.mission_failed_to_resume,
+            event=events.action_requests.resume_mission.failure,
             handler=_lockdown_mission_failed_to_resume,
         ),
         EventHandlerMapping[EmptyMessage](
-            event=events.robot_service_events.mission_succeeded,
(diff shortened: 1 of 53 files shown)


src/isar/state_machine/states/stopping_go_to_recharge.py
```diff
@@ -8,22 +8,26 @@
...
 
+    def _mission_stopped_event_handler(
+        aborted_mission: AbortedMission | EmptyMessage,
+    ) -> Transition:
+        if isinstance(aborted_mission, AbortedMission):
+            return GoingToRechargingWithMission.transition_and_start_return_home(
+                aborted_mission
+            )
+        else:
+            return GoingToRecharging.transition_and_start_return_home()
```
andchiind, line 19: I really wish Python had native pattern matching syntax 🫤

src/isar/state_machine/states/going_to_lockdown_with_mission.py
```diff
@@ -31,15 +31,15 @@ def _lockdown_mission_failed_to_resume(_: EmptyMessage) -> Transition:
...
             handler=_lockdown_mission_failed,
         ),
         EventHandlerMapping[EmptyMessage](
-            event=events.robot_service_events.mission_failed_to_resume,
+            event=events.action_requests.resume_mission.failure,
             handler=_lockdown_mission_failed_to_resume,
         ),
         EventHandlerMapping[EmptyMessage](
-            event=events.robot_service_events.mission_succeeded,
+            event=events.action_requests.execute_mission.success,
```
oysand, line 42: If execute mission always has success and failure. Does it then make sense to join the event handlers as well?

andchiind: I think not since they have different transitions. That way the transition and event becomes more descriptive, which is helpful for testing and logging.

oysand: Agreed

oysand (approved):
LGTM

PR #1196, opened 2026-09-10
Simplify models

### LINKED ISSUE #1194: Align start mission definition and mission model
There is not really a need to have separate StartMissionDefinition and Mission models, since the contents within them is the same. The only differences are minor naming differences, but they are otherwise equivalent. In the past the IDs would be different, but that is no longer the case.

It will reduce the amount of logic to maintain. Currently there are a lot of functions dedicated to translating these two models, as well as tests to back them up. Whenever a minor change to the models is made, a lot of code needs to be updated. As of the making of this issue, a lot of the example models used in tests do not align with the actual models, since many of the fields have been deleted, and people forgot to update the example JSON. Even if we end up having two different models in the end, we would still benefit from aligning the fields that we can, for instance the pose models.

Examples:
Image

Image

Image

andchiind:
Closes #1194 

See https://github.com/equinor/flotilla/pull/2984

Draft until I make PRs in ISAR implementations for return home change.

- [x] A test has been written

### DIFF
--- a/src/isar/robot/robot_monitor_mission.py
+++ b/src/isar/robot/robot_monitor_mission.py
@@ -157,12 +157,53 @@ async def get_and_report_task_status(
         return None
 
 
+async def robot_monitor_return_home_mission(
+    mission_id: str,
+    robot: RobotInterface,
+) -> tuple[ErrorMessage | None, bool]:
+    logger = logging.getLogger("robot")
+    logger.info("Started monitoring return home mission")
+    error_message: ErrorMessage | None = None
+
+    mission_status: MissionStatus = MissionStatus.InProgress
+    try:
+        while mission_status in [
+            MissionStatus.NotStarted,
+            MissionStatus.InProgress,
+            MissionStatus.Paused,
+        ]:
+            try:
+                mission_status = await get_mission_status(robot, mission_id)
+            except RobotMissionStatusException as e:
+                logger.exception("Failed to collect mission status")
+                error_message = ErrorMessage(
+                    error_reason=e.error_reason,
+                    error_description=e.error_description,
+                )
+                mission_status = MissionStatus.Failed
+
+            await asyncio.sleep(settings.FSM_SLEEP_TIME)
+
+        if error_message is None and mission_status in [
+            MissionStatus.Failed,
+            MissionStatus.Cancelled,
+        ]:
+            error_message = ErrorMessage(
+                error_reason=None,
+                error_description="The mission was reported as failed by the robot",
+            )
+        return error_message, False
+    except asyncio.CancelledError:
+        return None, True
+    finally:
+        logger.info("Stopped monitoring return home mission")
+
+
 async def robot_monitor_mission(
     mission: Mission,
     robot: RobotInterface,
     request_inspection_upload: Callable[[InspectionTask], None],
     mqtt_queue: MQTTQueue,
-    should_report_task_status: bool,
 ) -> tuple[ErrorMessage | None, Mission, bool]:
     logger = logging.getLogger("robot")
     logger.info(f"Started monitoring mission {mission.name}")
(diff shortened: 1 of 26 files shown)


src/isar/robot/robot_monitor_mission.py
```diff
@@ -157,12 +157,53 @@ async def get_and_report_task_status(
         return None
 
 
+async def robot_monitor_return_home_mission(
```
andchiind, line 160: I could move this to a different file, but I felt it was convenient to have here since it uses the get_mission_status function. I don't mind moving it though if people disagree.

src/isar/robot/robot_return_home.py
```diff
@@ -0,0 +1,50 @@
...
+    ErrorMessage,
+    ErrorReason,
+    RobotAlreadyHomeException,
+    RobotException,
+    RobotInfeasibleMissionException,
+)
+from robot_interface.robot_interface import RobotInterface
+
+
+def robot_return_home(
```
andchiind, line 14: At some point we could consider making a common approach to calling the robot interface functions, but for now I don't think some similar code is a problem.

src/isar/robot/robot_service.py
```diff
@@ -53,22 +57,34 @@ def stop(self) -> None:
         self.status_thread = None
         self.battery_thread = None
 
-    def _start_mission_handler(self, mission: Mission) -> bool:
-        error_message: ErrorMessage | None = robot_start_mission(
-            self.signal_exit, self.robot, self.logger, mission
+    def _start_return_home_handler(self, mission_id: str) -> bool:
```
andchiind, line 60: We store the mission_id since it is used when getting the mission status. We could consider removing this and just fetching the mission status from the current mission, but I wanted to limit breaking changes in this PR.

src/isar/robot/robot_service.py
```diff
@@ -136,17 +152,37 @@ def _resume_mission_handler(self) -> None:
         else:
             self.action_requests.resume_mission.trigger_success_response(EmptyMessage())
 
-    async def _monitor_mission_handler(self, mission: Mission) -> Mission | None:
-        remaining_mission: Mission | None = None
+    async def _monitor_return_home_handler(self, mission_id: str) -> None:
```
andchiind, line 155: I felt it was cleaner to make a new function as opposed to trying to combine it with the other monitor mission handler

src/isar/robot/robot_service.py
```diff
@@ -184,6 +220,18 @@ async def _run_main_event_loop(self) -> None:
         monitor_mission_task: asyncio.Task[AbortedMission | None] | None = None
 
         while not self.signal_exit.is_set():
+
+            return_home_request = (
+                self.action_requests.return_home.request.consume_event()
+            )
+            if return_home_request:
+                return_home_mission_id = str(uuid4())
```
andchiind, line 228: The ID doesn't matter, it just needs to be consistent when getting the mission status later

src/isar/state_machine/states/going_to_lockdown.py
```diff
@@ -5,14 +5,13 @@
...
 from robot_interface.models.exceptions.robot_exceptions import ErrorMessage
-from robot_interface.models.mission.mission import ReturnHomeMission
 
 
 def GoingToLockdown(events: Events) -> State:
 
     event_handlers: list[EventHandlerMapping] = [
         EventHandlerMapping[ErrorMessage](
-            event=events.action_requests.execute_mission.failure,
+            event=events.action_requests.return_home.failure,
```
andchiind, line 14: I am very happy with this change. It makes it much more clear what event we are actually expecting here.

src/robot_interface/models/mission/task.py
```diff
@@ -53,18 +53,15 @@ class ZoomDescription(BaseModel):
     objectHeight: float
 
 
-class Task(BaseModel):
+class InspectionTask(BaseModel):
```
andchiind, line 56: I could have kept using the name "Task", but "InspectionTask" is referred to in some ISAR implementations, so I kept that instead to avoid more breaking changes.

PR #1199, opened 2026-09-16
Allow setting the remaining return home timeout

### LINKED ISSUE #1198: Allow for extending the timeout before robot start return home mission.
We would like to be able to extend the waiting period before the robot returns home. This could for instance be done through an api call to extend the time by a given amount. 

When the battery is low, we still want the robot to abort and go home to avoid the robot running out of battery in field.

Christdej:
## Summary

Add `POST /schedule/set-return-home-timeout` with a body such as `{"seconds": 60}`. The command replaces the remaining wait from when the state machine handles the request, allowing the timeout to be shortened or extended.

- Accept positive integer seconds only while in `AwaitNextMission`, using the existing authenticated API and request/response event flow.
- Reset the active timer without transitioning states or changing the configured default for future waiting periods. Use a monotonic clock for elapsed time.
- Preserve low-battery return home, with pending battery events handled before timeout updates.

Closes #1198.

## Validation

- Targeted scheduler API, scheduling utilities, and waiting-state tests: 45 passed.
- State-machine test suite: 85 passed.
- Type checks, Ruff, isort, Black, and diff whitespace checks passed for affected files.
- Regression coverage includes shorter and longer waits, repeated updates, unrelated timers, default preservation, low-battery priority, invalid inputs, and state conflicts.

- [x] A test has been written


eivindsjovold (approved):
lgtm

